PR #170, opened 2015-01-05
Improve rfc6979 compatibility

swansontec:
The previous implementation didn't provide any way to try a different
nonce if the signature is invalid.

### DIFF
--- a/src/math/ec_keys.cpp
+++ b/src/math/ec_keys.cpp
@@ -100,7 +100,41 @@ bool verify_private_key(const ec_secret& private_key)
     return secp256k1_ec_seckey_verify(private_key.data()) == 1;
 }
 
-ec_secret create_nonce(ec_secret secret, hash_digest hash)
+endorsement sign(ec_secret secret, hash_digest hash)
+{
+    init.init();
+
+    int out_size = max_endorsement_size;
+    endorsement signature(out_size);
+
+    ec_secret nonce;
+    unsigned index = 0;
+    do {
+        nonce = create_nonce(secret, hash, index++);
+    } while (secp256k1_ecdsa_sign(hash.data(), hash.size(), signature.data(),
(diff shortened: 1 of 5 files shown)


src/math/ec_keys.cpp
```diff
@@ -128,7 +162,7 @@ ec_secret create_nonce(ec_secret secret, hash_digest hash)
     {
         V = hmac_sha256_hash(V, K);
 
-        if (verify_private_key(V))
+        if (0 == index--)
```
dcousens: IMHO, `index == 0` and then `index -= 1` would save everyone the hassle of having to remember prefix/postfix operators.

swansontec: Maybe, but unless you are a Python developer, the `++` and `--` operators are a perfectly common, everyday operators. I think we can assume that people reading this code are fairly familiar with the language. Otherwise, I suppose we could do this:

    if (0 == index)
        return V;
    --index;

dcousens: It's the mixing of mutation and comparison that I feel is confusing, not
necessarily the use of the postfix op in general.

On 6 Jan 2015 10:03 am, "William Swanson" <notifications@github.com> wrote:
>
> In src/math/ec_keys.cpp:
>
> > @@ -128,7 +162,7 @@ ec_secret create_nonce(ec_secret secret,
hash_digest hash)
> >      {
> >          V = hmac_sha256_hash(V, K);
> >
> > -        if (verify_private_key(V))
> > +        if (0 == index--)
>
> Maybe, but unless you are a Python developer, the ++ and -- operators are
a perfectly common, everyday operators. I think we can assume that people
reading this code are fairly familiar with the language. Otherwise, I
suppose we could do this:

swansontec: Oh, that makes sense. Changed and pushed.

PR #172, opened 2015-01-07
Allow --with-libsecp256k1 parameter

PabloCastellano:
./configure --with-libsecp256k1=/path/to/lib

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -54,6 +54,18 @@ AC_ARG_WITH([pkgconfigdir], AS_HELP_STRING([--with-pkgconfigdir=PATH],
     [pkgconfigdir="$withval"], [pkgconfigdir='${libdir}/pkgconfig'])
 AC_SUBST([pkgconfigdir])
 
+# Allow libsecp256k1 to be installed in a custom path:
+AC_ARG_WITH([libsecp256k1],
+            AS_HELP_STRING([--with-libsecp256k1],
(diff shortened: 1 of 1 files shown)


evoskuil:
The `PKG_CHECK_MODULES([libsecp256k1], [libsecp256k1])` line should be setting the proper flags and those values should be referenced in the MakeFile: https://github.com/libbitcoin/libbitcoin/blob/version1/src/Makefile.am#L4

The problem appears to be that the MakeFile is not referencing the lib values: https://github.com/libbitcoin/libbitcoin/blob/version1/src/Makefile.am#L57

See the version2 MakeFile for reference: https://github.com/libbitcoin/libbitcoin/blob/version2/Makefile.am#L244

pmienk:
Please note if you're installing libsecp256k1 in an alternate location, you can pass the appropriate information via `PKG_CONFIG_PATH` at configure time via `PKG_CONFIG_PATH=<path> ./configure`.  Note that by default on Ubuntu `PKG_CONFIG_PATH=/usr/local/lib/pkgconfig`.  The existence of this mechanism makes the changes provided unnecessary to accomplish an alternate installation location.

PabloCastellano:
Hi. I was just trying to fix install-sx.sh. It fails compiling libbitcoin when you don't install it in the default path. Closing PR and will try to fix it properly. Thanks

PR #176, opened 2015-01-08
added sign byte back to stealth.

genjix:
This increases the number of client side computations, which reduces the
anonymity. It is trivial for the server to store this data and halves
the computation workload for the client, thereby increasing the
acceptable working dataset & improving anonymity for the client.
In fetch_stealth it returns rows with the ephemeral key. The beginning sign byte (02
and 03) was remove and have the client test both combinations.
On further reflection after speaking with caedes, it is a bad idea and
reduces anonymity.
We want to lessen the processing burden on clients, so they can use a
smaller prefix (= more data) to remain more anonymous.


veox:
[Discussion](https://lists.dyne.org/lurker/message/20150108.041153.9d561dfa.en.html) on ML.

genjix:
Rejected. Instead we will use 02 instead.

PR #178, opened 2015-01-14
Fix race condition in channel

genjix:
Basically the stopped_ variable exists to avoid code which closes the socket and calls the handlers from being called twice.

Since the stop_impl() code may be called by different threads, it can happen that the code gets executed eventhough the check at the beginning [if (stopped_) return...] passes. This fixes that by adding a mutex.

### DIFF
--- a/include/bitcoin/bitcoin/network/channel.hpp
+++ b/include/bitcoin/bitcoin/network/channel.hpp
@@ -248,7 +248,7 @@ private:
     void stop_impl();
     void clear_subscriptions();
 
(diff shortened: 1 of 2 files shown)


evoskuil:
Does this fix any of the outstanding libbitcoin-server issues?

genjix:
No, it is unrelated.

genjix:
There's problems with this commit. It's stopping the server to exit. For some reason the lock can never be acquired.

PR #180, opened 2015-01-14
remove the mutex, use the strand instead.

genjix:
Instead of using a mutex. I instead use the strand. This seems to work better.


evoskuil:
Could you rebase this to master? I merged the revert first and that broke this merge.

genjix:
ok! done.

PR #187, opened 2015-01-23
ENABLE_TESTNET propagation through pc.in.

evoskuil:
Thanks @genjix for the suggestion. This guarantees consistency with dependency builds and simplifies their configure.ac.


genjix:
excellent, can't say I know completely about autoconf, but from my limited knowledge it looks good.

evoskuil:
I verified it with a positive and negative run that the below value is conditionally set. As long as the dependencies pick this up from package config it should be good:

```
Cflags: -I${includedir} -std=c++11 -DENABLE_TESTNET
```

PR #188, opened 2015-01-23
Limit boost namespace and type defs in config.

evoskuil:


evoskuil:
This should make master backward compatible with version2.

PR #207, opened 2015-03-05
Adds BIP39 mnemonic support

thecodefactory:
BIP39 support includes creation from entropy and from mnemonic words to seed, as well as appropriate unit tests.


evoskuil:
Rebased in https://github.com/libbitcoin/libbitcoin/pull/210

PR #224, opened 2015-03-25
missing m4 for autoconf detection of boost_chrono libs

jaromil:
the file is part of GNU autoconf, taken from https://www.gnu.org/software/autoconf-archive/ax_boost_chrono.html


evoskuil:
Wondering why we didn't see Travis failures without this.

jaromil:
Perhaps because you are using libbitcoin-build in Travis and that has this .m4

evoskuil:
Libbitcoin-build is used to generate the build artifacts which are then checked in. It's not run as part of the  build of another repo. (Eventually libbitcoin-build will distribute the m4 scripts as well, which would have caught this drift.)

PR #227, opened 2015-03-28
Mnemonic optimization

swansontec:
I literally could not compile the mnemonic code on some of our boxes, since the compiler needed more RAM than was available. The problem isn't the raw data; it's the thousands of std::string constructor calls that the compiler has to emit. Besides taking a ton of time and memory to compile, these constructors now need to run each time the library boots up, adding run-time overhead.

This code is slightly less pretty, since it uses pointers instead of nicely-nested STL types, but I think the performance gains are well worth it. Being able to compile the code at all is nice, too.


thecodefactory:
Just curious what kind of machine you're using.  I ask because I'm responsible for that and didn't realize there was an issue as I'm able to build properly on a laptop with 4GB RAM in a reasonable amount of time.  That said, I'm all for an in improvement because I agree that std::string is not the best type for that usage.

swansontec:
We use cheap VPS boxes as libbitcoin servers, and the goal is to pay for as little RAM as possible. I also run bx on a Raspberry Pi for personal use.

thecodefactory:
Makes sense.  Carry on, and thanks!

swansontec:
Oops, I had no idea that ndebug was enabled by default. I've fixed the assert and re-pushed.

PR #234, opened 2015-04-05
Mnemonic clean-up

swansontec:
After a thorough audit, I think this code is now ready for prime-time.

There were a few places where the code was violating the bip39 standard, so I have fixed those. I have also removed the `language` enum. For a detailed rationale, see the `Allow user-defined mnemonic wordlists` commit notes. Finally, I have removed the unnecessary bip39 namespace, as per yesterday's discussion.

The final commit in this series renames "dictionary" to "wordlist", to better match the contents. I can't run the libbitcoin_build tool to freshen the headers, so I just did a search & replace. I would much prefer to do this the right way, so let's hold off on merging until libbitcoin_build is working again.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -83,7 +83,7 @@ src_libbitcoin_la_SOURCES = \
     src/utility/threadpool.cpp \
     src/utility/timed_section.cpp \
     src/wallet/address.cpp \
-    src/wallet/dictionary.cpp \
+    src/wallet/wordlist.cpp \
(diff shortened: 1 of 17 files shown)


swansontec:
Actually, now that I'm working on the libbitcoin-explorer integration, it seems like I should probably include the language string inside the wordlist map. That would make lookups a lot easier. So, let's hold off on merging this until I fix that.

Given the holiday, I may not get back to working on this until Monday.

PR #236, opened 2015-04-05
Mnemonic clean-up, second try

swansontec:
This is an update on pull request #234. Here are the fixes:

* Remove the unnecessary pbkdf2 assertions.
* Expose the built-in dictionaries as a std::map for run-time string-based lookup.
* Rename `wordlist` to `dictionary`, as requested.

I expose the built-in dictionaries two ways, once as plain C++ variables and again through a string-based std::map. If a developer knows which language they want at compile-time, they can simply refer to the plain variables. This removes the indirection of an enum, and makes the compile-time language lookup a lot simpler. Its main users are default parameters and unit tests, so I would be willing to privatize this capability if necessary.

For run-time language lookup, the developer can examine the map to determine which languages are available. The libbitcoin/libbitcoin-explorer#231 pull request makes use of this to eliminate the need for switch statements.

### DIFF
--- a/include/bitcoin/bitcoin/math/hash.hpp
+++ b/include/bitcoin/bitcoin/math/hash.hpp
@@ -101,8 +101,8 @@ BC_API long_hash hmac_sha512_hash(data_slice data, data_slice key);
  *
  * pkcs5_pbkdf2_hmac_sha512(passphrase, salt, iterations)
  */
-BC_API bool pkcs5_pbkdf2_hmac_sha512(const std::string& passphrase,
-    data_slice salt, size_t iterations, long_hash& long_hash);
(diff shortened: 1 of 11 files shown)


evoskuil:
I'm in favor of these changes with the exception of the weakly-typed dictionary identifier enumeration. See comments in the corresponding bx pull request: https://github.com/libbitcoin/libbitcoin-explorer/pull/231.

swansontec:
Ok, I have put all the dictionaries into their own namespace. I couldn't make a `dictionary` namespace, since that's already a type name. Instead, I put them in the `langauge` namespace, which matches the name of the old enum.

PR #242, opened 2015-04-21
Fix mnemonic passphrase normalization.

evoskuil:


evoskuil:
This is on hold for a day or two while ICU integration is getting wrapped up.

PR #266, opened 2015-06-15
- pull in necessary changes from master to allow explorer to build

thecodefactory:
explorer doesn't build against version2 after recent changes.  these are the minimal changes needed to get it building again.


evoskuil:
Thanks, I didn't realize it was so much. I probably should have rolled back bx at the same time as the server stack.

PR #274, opened 2015-07-19
Update of configure.ac to work with OpenBSD 5.7 / Autotools 2.69.

mlmikael:
Update of configure.ac to work with OpenBSD 5.7 / Autotools 2.69.

"AM_PROG_AR" was introduced with some newer Autotools version than 2.69, which is the newest one available as a package for OBSD. Similar case with "AM_PROG_CC_C_O".


mlmikael:
@evoskuil,

] Feel free to make a pull requests against the doc(s).
Done.

] Would be nice if would could catch it in the build, but I'm not sure how yet.
Me neither.

evoskuil:
Cherry-picked readme and incorporated generated version comment: https://github.com/libbitcoin/libbitcoin/pull/277

PR #294, opened 2015-08-18
Fixed pubkey script payment type checking

waldanbridger:
Fixed NULL_DATA check function.
Bare MULTISIG (bip11) tx out check implemented.


src/script.cpp
```diff
@@ -1500,11 +1500,27 @@ bool is_stealth_info_type(const operation_stack& ops)
     return ops.size() == 2 &&
         ops[0].code == opcode::return_ &&
         ops[1].code == opcode::special &&
-        ops[1].data.size() >= hash_size;
+        ops[1].data.size() <= 40;
```
evoskuil: Hi @waldanbridger, could you explain this change?

waldanbridger: From the [official developer guide](https://bitcoin.org/en/developer-guide#null-data):
> **Null Data (OP_RETURN) Transaction**
> Pubkey Script: OP_RETURN <0 to 40 bytes of data>

Also, some pools (e.g.[eligius](http://eligius.st/~gateway/faq-page)) allows up to 80 bytes of data:
> Data carrier transactions are permitted to be up to 80 bytes.

src/script.cpp
```diff
@@ -1531,11 +1553,24 @@ bool is_script_code_sig_type(const operation_stack& ops)
         return false;
     }
     const operation_stack& code_ops = script_code.operations();
-    // Minimum size is 4
+    const payment_type redeem_type = script_code.type();
     // M [SIG]... N checkmultisig
+    // or one of the other standard transaction types
     return code_ops.size() >= 4 &&
```
waldanbridger: [BIP16](https://github.com/bitcoin/bips/blob/master/bip-0016.mediawiki#Specification) says:
> Transactions that redeem these pay-to-script outpoints are only considered standard if the serialized script - also referred to as the redeemScript - is, itself, *one of the other standard transaction types*.

src/script.cpp
evoskuil: It's sometimes hazardous to rely on documentation. The [default maximum OP_RETURN in the satoshi client](https://github.com/bitcoin/bitcoin/pull/5286) is 80 bytes.

The documented [stealth transaction format](https://wiki.unsystem.net/en/index.php/DarkWallet/Stealth) indicates that 38 bytes is required:

```
RETURN <version:1=6> <nonce:4> <P:33>
```

However this documentation is also dated.

The version byte was dropped, the nonce was moved to the right of the public key, the public key was restricted to 32 bytes (sign fixed by convention), and the nonce length was made variable (any number of bytes up to 80 at this point). This was done to allow the value to match that of more arbitrary data in the blockchain by not being fixed. It also served to increase nonce space. I believe this change was documented but I can't locate it presently.

So the proper check for length of a stealth `OP_RETURN is >= hash_size` (32 bytes). The `read_ephemkey()` function assumes this symbol, so it's both a safety and readability issue.

src/script.cpp
evoskuil: Thanks for implementing this. I may not have time to review it myself in the next few days, but I'd like to get a review from some of the other maintainers, because it's in script.

src/script.cpp
waldanbridger: Okay, I see. So, should I implement a separate 'is_null_data_type' method for standard NULL_DATA type outputs? Because, as I understand, the stealth transaction format is a feature of libbitcoin and the original bitcoin-core has no idea about it. Am I right? Even if it's not, blockchain already contains many NULL_DATA outputs (where data < 32 bytes) and they must be handled properly, because they're standard.
But in case of separate is_null_data_type and is_stealth_info_type there will be a conflict because stealth_info_type is a subset of null_data_type.

waldanbridger:
>Okay, I see. So, should I implement a separate 'is_null_data_type' method for standard NULL_DATA type outputs? Because, as I understand, the stealth transaction format is a feature of libbitcoin and the original bitcoin-core has no idea about it. Am I right? Even if it's not, blockchain already contains many NULL_DATA outputs (where data < 32 bytes) and they must be handled properly, because they're standard.
But in case of separate is_null_data_type and is_stealth_info_type there will be a conflict because stealth_info_type is a subset of null_data_type.

Let's continue the conversation here, in comments. (stealth_info_type vs null_data_type)

src/chain/operation.cpp
```diff
@@ -194,7 +194,7 @@ std::string operation::to_string() const
     if (data.empty())
         ss << opcode_to_string(code);
     else
-        ss << "[ " << encode_base16(data) << " ]";
+        ss << "[" << encode_base16(data) << "]";
```
evoskuil: This change would require a corresponding change to the from string implementation (to support both formats). It would also constitute a breaking change in the case where API users depended on the format, although we can accept breaking changes in master (3.0.0).

evoskuil:
The reason stealth uses OP_RETURN is that there are no semantics associated with the output. It is an objective that stealth data be as opaque as possible. The `script_type::type()` implementation really should not return `payment_type::stealth_info` at all. `stealth_info` should not be part of the type enumeration. There should be an enumerated value for OP_RETURN scripts and the stealth test should be independent (i.e. applied as a subsequent test after confirming the script is OP_RETURN). And the test is by design simply a minimum length test. There is a trivial corresponding change in `libbitcoin-blockchain` but this is a simple set of changes, and we can accept the interface break in 3.0.0 so now is the time to do it.

src/chain/operation.cpp
```diff
@@ -329,11 +352,25 @@ bool is_script_code_sig_type(const operation::stack& ops)
...
+    // or one of the other standard transaction types
     return script_code.operations.size() >= 4 &&
         count_non_push(script_code.operations) == 1 &&
-        script_code.operations.back().code == opcode::checkmultisig;
+        script_code.operations.back().code == opcode::checkmultisig ||
+        redeem_type == payment_type::pubkey;
+        redeem_type == payment_type::pubkey_hash;
+        redeem_type == payment_type::script_hash;
+        redeem_type == payment_type::null_data;
+}
```
evoskuil: This is probably not what is intended.

evoskuil:
@waldanbridger Please see https://github.com/libbitcoin/libbitcoin/pull/321

evoskuil:
These changes have been merged with https://github.com/libbitcoin/libbitcoin/pull/321. Thanks for the help!

PR #300, opened 2015-08-27
Changes for bip-38 in alternate build configurations

thecodefactory:
This allows explorer to properly utilize subsets of bip38 when icu is not enabled (lock intermediate mode only), or testnet is enabled (no tests yet).


evoskuil:
Oops, I lost track of this. t may be too late for most of it at this point.

thecodefactory:
No, it should still apply, the tests in libbitcoin master are still broken.

evoskuil:
I'm in the process of extracting the hard chain dependency which only exists because of the address version byte.

evoskuil:
This has become obsolete due to changes implemented to support altchains without the need for conditional by chain.

PR #313, opened 2015-09-16
Fix base16 locale dependency issues

swansontec:
The C standard library character classification functions depend
on the current locale, and can give crazy answers in some situations.

Provide our own is_base16 function that unambiguously corresponds
to the valid character range for that format.


evoskuil:
Well that's behavior I didn't expect - thanks.

evoskuil:
Do you need this backported back to `version2`?

swansontec:
No, it's more of a "huh, that's weird" thing.

PR #314, opened 2015-09-17
New general-purpose URI parsing module

swansontec:
This exists to help with BitID and other things of that nature. The code works, and I've tried to stick to the latest coding standards, but this could probably use good solid review before accepting.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -164,6 +164,7 @@ src_libbitcoin_la_SOURCES = \
     src/wallet/select_outputs.cpp \
     src/wallet/stealth.cpp \
     src/wallet/stealth_address.cpp \
+    src/wallet/uri.cpp \
     src/wallet/wif_keys.cpp \
     src/wallet/parse_encrypted_keys/parse_ek_key.hpp \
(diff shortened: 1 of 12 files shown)


swansontec:
Also see https://github.com/swansontec/libbitcoin/pull/1 for work that builds on top of this.

swansontec:
Ah, it looks like `encode` should be called `to_string`. Didn't realize that that was a project-wide change.

evoskuil:
Thanks @swansontec, this looks nice.

evoskuil:
@swansontec There's a lot of change right now in wallet. Don't worry about `to_string` - That may end up being limited to the satoshi and blockchain primitives, since `encode` would be ambiguous for them.

I'm using encode/decode for each of the `base58` wallet types, with intermediate array/chunk types and parsed struct/class types. The deserialization exceptions will end up going away for the wallet types as well. I've written `bool` validity operator overrides for each of their class/struct forms. There will be a pull request on this soon.

It's a work in progress, heading towards a more rational view of all primitives, while retaining the core philosophy.

PR #317, opened 2015-09-19
bitcoin_uri reader-writer based on new interface/parser

evoskuil:
This creates a unified reader-writer based on the scheme-agnostic `uri`.

The `uri_parse` function has been templatized to resolve the problem of returning a partially-initialized URI in the failure case. It has also been moved statically into `uri_reader`, as it accepts a reader as its template (formerly out) parameter.

Bitcoin semantics have been removed from the general purpose parse function by incorporating them into the visitor (implemented by the `bitcoin_uri`). The `uri_reader` interface is expanded to require full general  parsing callbacks, allowing it to support any uri scheme.

The unified reader-writer hides all properties behind accessors in order to maintain consistent state. This was easier without using optional return values, and is generally easier to access, but it removes the ability for a user of `bitcoin_uri` to distinguish between not-set parameters and explicitly-provided parameters with default values. However in the Bitcoin uri scheme there is no meaningful difference, so hopefully this is sufficient.

### DIFF
--- a/include/bitcoin/bitcoin/wallet/bitcoin_uri.hpp
+++ b/include/bitcoin/bitcoin/wallet/bitcoin_uri.hpp
@@ -21,81 +21,66 @@
 #define LIBBITCOIN_WALLET_BITCOIN_URI_HPP
 
 #include <cstdint>
+#include <map>
 #include <string>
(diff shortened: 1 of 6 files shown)


skaht:
Cool...

PR #320, opened 2015-09-24
Remove ENABLE_TESTNET condition

evoskuil:
This is a significant refactor of wallet primitives. There are several related objectives.

These are all resolved:

* https://github.com/libbitcoin/libbitcoin/issues/308
* https://github.com/libbitcoin/libbitcoin/issues/315
* https://github.com/libbitcoin/libbitcoin/issues/319
* https://github.com/libbitcoin/libbitcoin/issues/318

This last item is not yet fully dialed in because of backward compatibility for existing stealth addresses, but the work is done.

There is a new `elliptic_curve` function `compress(const ec_uncompressed&)`

There are two new primitives, `ec_public` and `ec_private` which are layers over `ec_secret` and `ec_compressed` / `ec_uncompressed`. The `ec_point` blob is replaced with `ec_public`, which can emit as either compressed or uncompressed, and stores internally as compressed.

The implementation rationalizes WIF with `ec_private` as just a serialization, analogous to `payment_address` and HD serialization in base58 form. The `ec_private` and `ec_public` carry contextual information for chain (version), but there are still some gaps in this area because of the lack of standard `ec_public` WIF encoding. Public keys have no standard base58 serialization, so currently `ec_public` relies on base16. There is currently some discussion on a public WIF encoding, which would be useful here.

Stealth code has been shuffled to align with payment address and ec math functions.

All wallet dependency has been removed from the `chain` namespace, which results in a slight amount of redundant code.


evoskuil:
This should not be merged until related updates are applied to downstream repos.

PR #331, opened 2015-10-24
Networking redesign.

evoskuil:
This is a full rewrite of the `bc::network` namespace. The pattern remains proactor and the architecture is not materially changed. However code is factored into manageable units designed for test.


evoskuil:
Pushed to `network` branch.

PR #333, opened 2015-11-02
Speed up the payment_address hash function

swansontec:
The Airbitz wallet is seeing bad performance when we use bc::payment_address
with std::unordered_map. This change speeds things up about 3x.


evoskuil:
Nice

PR #334, opened 2015-11-09
Special opcode deserialization update

thecodefactory:
This is a subtle change that needs review.

Fixes libbitcoin/libbitcoin-server#134

With this applied, my testnet blockchain is currently at 508716 and counting.


thecodefactory:
Actually, this is incorrect, but strangely appears to work (this method returning false on special opcode handling in particular).  I'm going to close it for now and re-visit for a better fix.

PR #349, opened 2015-12-15
install.sh: OpenBSD support to CPU cores check

mlmikael:


evoskuil:
Incorporated into https://github.com/libbitcoin/libbitcoin/pull/357

PR #369, opened 2016-03-04
Re-organization and cleanups of png and qrcode additions

thecodefactory:


include/bitcoin/bitcoin/wallet/qrcode.hpp
```diff
@@ -29,60 +29,53 @@
...
-    const qr_case_sensitive cs, std::ostream& out);
-
-#endif
+class BC_API qr
+{
+    typedef QRencodeMode encode_mode;
+    typedef QRecLevel error_recovery_level;
+
+  public:
+    static constexpr uint32_t version = 0;
```
evoskuil: Headers need to use `BC_CONSTEXPR` vs. `constexpr`. This allows library linkage into VC++ 12.0 compiled builds (without CTP upgrade). This is a requirement for working with QT in MSVC.

src/wallet/qrcode.cpp
```diff
@@ -27,46 +27,44 @@
...
     data_sink ostream(out);
     data_source istream(data);
 
-    if (qrencode_data(istream, version, level, hint, cs, ostream))
+    if (qr::encode(istream, version, level, mode, case_sensitive, ostream))
         return out;
 
-    return data_chunk();
+    return data_chunk{};
 }
```
evoskuil: Can be simply:
```cpp
return {};
```

evoskuil:
Looks great! Can you please make the `constexpr` change and rebase on the latest commits?

PR #381, opened 2016-03-25
Updated Doxygen file to generate docs

shawnpringle:
The Doxyfile seems to be one that has just been forgotten about.  Changes make Doxygen recurse so even if the file continues to be ignored new directories will be incorporated into the generated documentation.


evoskuil:
Yes, there has been no attention paid to this in some time. Thanks for looking into it.

PR #382, opened 2016-03-26
Updates build_from_tarball to respect conditional build arguments

### LINKED ISSUE #379: All builds always build icu, png and qrencode.
All repos, all builds, are now always building these three dependencies even if they are not selected. This is a regression where the build condition was removed guring tarball build generalization.

thecodefactory:
Resolves #379


evoskuil:
I've just completed a rework of the installer, which obsoletes these changes.

PR #406, opened 2016-06-21
resource_lock class to replace boost::interprocess::file_lock in libbitcoin-database use.

RojavaCrypto:


include/bitcoin/bitcoin/utility/resource_lock.hpp
```diff
@@ -0,0 +1,56 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_FILE_LOCK_HPP
+#define LIBBITCOIN_FILE_LOCK_HPP
```
evoskuil: Match file name here.

include/bitcoin/bitcoin/utility/resource_lock.hpp
```diff
@@ -0,0 +1,56 @@
...
+
```
evoskuil: #include <memory>

include/bitcoin/bitcoin/utility/resource_lock.hpp
```diff
@@ -0,0 +1,56 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/interprocess/sync/file_lock.hpp>
+
+namespace libbitcoin {
+
+typedef boost::filesystem::path file_path;
```
evoskuil: Avoid making this a global bc typedef. If necessary, define in the context of this class (as "path", by existing convention) but preferably just use the fully qualified name in the header and use a local typedef in th cpp file.

include/bitcoin/bitcoin/utility/resource_lock.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * A resource lock device that ensures exclusive access to a resource.
+ * It takes a path for creating a lock file object that can be used
+ * for telling the usage of some resource between processes.
+ * Example: opening/closing blockchain database.
+ */
+class resource_lock
```
evoskuil: resource is overly general, could this be process_lock or file_lock?

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <bitcoin/bitcoin/utility/resource_lock.hpp>
+
+#include <boost/filesystem.hpp>
```
evoskuil: #include <memory>

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <bitcoin/bitcoin/utility/resource_lock.hpp>
+
```
evoskuil: #include <memory>

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include <boost/filesystem.hpp>
+#include <boost/interprocess/exceptions.hpp>
+#include <bitcoin/bitcoin/unicode/ifstream.hpp>
+#include <bitcoin/bitcoin/unicode/ofstream.hpp>
+#include <bitcoin/bitcoin/utility/assert.hpp>
+
+namespace libbitcoin {
+
+void touch_file(const file_path& path)
```
evoskuil: Prefer private static method to file scope function. This facilitates testing of the method (in which case it would need to become protected) and more strongly organizes code.

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+    try
+    {
+        lock_ = std::make_shared<boost_file_lock>(lock_path_.c_str());
+        return lock_->try_lock();
+    }
+    catch (const boost::interprocess::interprocess_exception&)
+    {
+        lock_ = nullptr;
+        return false;
+    }
```
evoskuil: Skip line (following scope close, and preceding comment).

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+    // Should not arrive here!
+    BITCOIN_ASSERT_MSG(false, "Reached unreachable code path.");
+    return false;
+}
```
evoskuil: Skip line (following scope close).

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+bool resource_lock::unlock()
+{
+    boost::system::error_code ec;
+    boost::filesystem::remove(lock_path_, ec);
+    if (ec)
+        return false;
+    return true;
```
evoskuil: prefer simpler:

return !ec;

test/utility/resource_lock.cpp
```diff
@@ -0,0 +1,43 @@
...
+ */
+#include <boost/test/unit_test.hpp>
+
+#include <bitcoin/bitcoin.hpp>
+
+using namespace bc;
+
+BOOST_AUTO_TEST_SUITE(resource_lock_tests)
+
+BOOST_AUTO_TEST_CASE(resource_lock__duplicate_locks)
```
evoskuil: A negative test is also required in order to achieve full coverage.

Test naming convention is:

[class|file__method|function__condition_under_test__expectation]

evoskuil:
Good add, I think we had a todo for this in blockchain, thanks. See style comments.

include/bitcoin/bitcoin.hpp
```diff
@@ -112,6 +112,7 @@
 #include <bitcoin/bitcoin/utility/enable_shared_from_base.hpp>
 #include <bitcoin/bitcoin/utility/endian.hpp>
 #include <bitcoin/bitcoin/utility/exceptions.hpp>
+#include <bitcoin/bitcoin/utility/resource_lock.hpp>
```
evoskuil: Sort order (also in Makefile.am). These are generated by libbitcoin-build using automatic sort, do it would get reordered on next generation.

evoskuil:
The project and filter files should be updated to avoid breaking the MSVC builds. These aren't yet generated by libbitcoin-build. We should soon have CI builds for these as well.

include/bitcoin/bitcoin/utility/resource_lock.hpp
RojavaCrypto: It was called file_lock before, but I tried to more accurately express what the purpose of this class. I have another class for process_lock so that name is a no go. resource expresses that some resources are being locked but file_lock is also possible.

src/utility/resource_lock.cpp
RojavaCrypto: It's in the main hpp now, no need to include it twice.

src/utility/resource_lock.cpp
RojavaCrypto: Both the constructor and try_lock() can throw. Unless you mean something else.

src/utility/resource_lock.cpp
RojavaCrypto: What do you mean?

test/utility/resource_lock.cpp
RojavaCrypto: What kind of negative test? The test opens a lock, and sees if another lock can acquire it. It should fail. I can't think of any other tests.

RojavaCrypto:
ok, thanks for all the suggestions. I've made most of your changes. Once I understand the others, I'll also make those. About the name, resource_lock is good but file_lock is also possible.

evoskuil:
Cool. The name should reasonably reflect the purpose. The class provides a cross process indication that the same binary has a protected resource open in another process. So I suppose resource_lock is fine despite the fact that we use it to indicate exclusivity of a file resource.

test/utility/resource_lock.cpp
evoskuil: One test should execute the non-exceptional path and another should exercise the exceptional path (catch block). Presumably there must be a test that does not open the lock but tries to acquire it. Otherwise there is code that is untested. Coveralls should indicate uncovered lines.

Also, I haven't looked, but my understanding is that the lock applies in the interprocess scenario. I'm not sure therefore that an in-process test is sufficient.

src/utility/resource_lock.cpp
evoskuil: The } without a trailing ; (or continuiation such as "catch") closes a scope. This should be followed with a blank line as a matter of style. I've standardized this in most places.

src/utility/resource_lock.cpp
evoskuil: Something else, see below.

include/bitcoin/bitcoin/utility/pid_lock.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE. See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program. If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_APPLICATION_LOCK_HPP
+#define LIBBITCOIN_APPLICATION_LOCK_HPP
```
evoskuil: Match file name.

include/bitcoin/bitcoin/utility/pid_lock.hpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include <string>
+#include <boost/filesystem.hpp>
+
+namespace libbitcoin {
+
+typedef boost::filesystem::path file_path;
```
evoskuil: Scope to class or use only in cpp. Don't want to proliferate ad-hoc types in the namespace.

src/utility/resource_lock.cpp
```diff
@@ -0,0 +1,80 @@
...
+        return lock_->try_lock();
+    }
+    catch (const boost::interprocess::interprocess_exception&)
+    {
+        lock_ = nullptr;
+        return false;
+    }
+    // Should not arrive here!
+    BITCOIN_ASSERT_MSG(false, "Reached unreachable code path.");
+    return false;
```
evoskuil: Above two lines are dead code, cannot be reached and therefore will never be tested. Should just remove.

src/utility/resource_lock.cpp
evoskuil: By "main hpp" do you mean resource_lock.hpp? If so that's okay, but I prefer to include each reference in the file where it issued. This makes it safe to remove from a file if no longer used locally. Determine the transitive closure of all includes is humanly impossible, I got tired of chasing unpredictable compiler errors downstream of a seemingly innocuous change.

evoskuil:
Im not clear on the purpose for pid_lock. Also there area few cross platform issues. I can help resolve these, but I'm not able to code for a couple more weeks.

RojavaCrypto:
Damn, I didn't mean to add those files. I've forgotten how to use git properly.

Oh well, might as well get those reviewed too (they weren't finished).

RojavaCrypto:
Nonetheless I made those changes for pid_lock too.

The purpose of pid_lock is so we cannot run the binary twice. It will go in the console libbitcoin-server. On POSIX, the standard for single run daemons is to create a named file in ```/var/run/``` with the process ID. This path should be configurable at the ```./configure && make``` steps.

So the idea is if people try to run ```bs``` when another instance is running, it will warn them and tell them they can use a switch to bypass this check. If I want to run multiple instances, then I will have to call ```--no-process-check``` or something.

RojavaCrypto:
For Windows, I've left 3 stubs which are:

- Headers needed.
- Get current process ID (```getpid()```)
- Check whether the given process ID exists and is running.

These should be relatively easy to fill in. The other task is making the path configurable by the build systems (currently hard coded to ```/var/tmp/``` which is the user non-root path for non-packaged installs).

evoskuil:
Twice against the same store?

RojavaCrypto:
The ```resource_lock``` is for the blockchain, and the ```pid_lock``` is for the daemon.

Example:

User tries to read the blockchain using a utility program when it is already open. Warning issued.

User tries to run multiple versions of the server (not an invalid operation) and gets warned.

Both errors are different. My motivation with creating the pid_lock is better feedback and a reporting mechanism to dialog user errors. See more about PID files here: https://unix.stackexchange.com/questions/12815/what-are-pid-and-lock-files-for

RojavaCrypto:
As an example of the differences in logic.

If there is a lock on the blockchain store when you try to use it, the program will expect you to delete it before it will work. This is because in the event of a crash, the blockchain will have been left in an undefined state so it should not work.

For the pid_lock, it will simply read the process ID in the file and test whether that process is running. If not, it will just ignore the fail because a program can close without deleting its PID file. We will probably put an assert though in the debug version to alert us when we aren't closing the pid_lock on program shutdown. But for the user they shouldn't need this.

evoskuil:
I cannot see a reason to warn or prevent concurrent use of the same source code against different blockchains. Concurrent use of one blockchain is already prohibited, and must remain so.

evoskuil:
I get the mechanics, I just cannot see a valid use case. PID files are used to prevent multiple instances of the same process. We have no reason to care about this if the processes are using different databases, and we are already prohibiting use of the same database using the resource lock.

RojavaCrypto:
Here's a practical example:

I have a program that provides a web interface for managing the server through a bunch of scripts. The PID file in ```/var/run/bs.pid``` tells me the process is running. Maybe I also have a bunch of scripts which are periodically running that also interact with the daemon (checking it hasn't crashed) or maybe they stop the daemon to make a backup of the blockchain. Or maybe I just want to make a kill_bs command which is ```kill $(cat /var/run/bs.pid)```. There are many uses for a PID file.

The difference between the resource_lock and pid_lock is that resource_lock is application specific, while pid_lock is general.

Sometimes there are additional programs and daemons which will run alongside the server, and they need a mechanism to see if it's running.

Obviously only one instance can create a PID file, so trying to run multiple instances means you have to be explicit about your intention, but in return you get better error reporting for users (you have multiple instances running) catching classes of bugs that can be caused by other problems (for instance, blockchain might or might not be in use by another bs instance - it's not clear).

evoskuil:
Also, I think your interpretation of the boost interprocess lock isn't quite accurate. The lock only holds if another process currently holds the lock. The file is in fact ignored if there is no process holding the associated interprocess lock.

evoskuil:
Database corruption detection needs to arise from a write indicator tfat is cleared on memory map flush to disk. If the indicator is left behind, the database is corrupt, and *any* subsequent attempted use must fail (not yet implemented).

RojavaCrypto:
```lock->try_lock()``` returns true on lock, false otherwise. If it returns false, we error.

evoskuil:
The blockchain database cannot be in use by multiple processes, so we should eliminate that use case.

There are no problems, logging or otherwise, with running multiple instance of a bs and/or bn process, unless the admin configures the same ports (bs) or database/log paths. Those aren't problems we want to prevent by limiting use.

So the use case that seems to be interesting is detection by a script to determine if a particular process is running. In my experience (and in the discussion on the link you sent) this is generally done using platform specific tooling, external to the process itself.

> ```lock->try_lock()``` returns true on lock, false otherwise. If it returns false, we error.

But it doesn't return true unless another running process holds a lock on the file.

RojavaCrypto:
PID files are completely standard. Tor, Apache, systemd, cups, nginx, ... have them. It's a good mechanism that we can see if another instance is running, or if another daemon is blocking the ports we need. The current messages when running multiple instances are cryptic.

If you want to run another instance, you need to change the blockchain path, the config files and all the ports. It's hardly another big hurdle to ask people to change or disable the PID path.

evoskuil:
> PID files are completely standard. Tor, Apache, systemd, cups, nginx, ... have them.

I have not suggested that it is nonstandard. My concern is that this is a solution for a problem we do not have, and it is not actually a solution to the issues raised.

> It's a good mechanism that we can see if another instance is running,
 
There has been no valid reason raised for our code to know if there is another instance of our code running, whether it is tied to a binary, like some version(s) of bs and/or bn, or arbitrary code compiled to access the blockchain database, logs, node ports, or server ports.

With this knowledge there are only two things our code could do: (1) terminate with a message, or (2) present a general warning and continue. It cannot fix misconfiguration and knowledge that another (related?) process is running is insufficient information to provide actionable feedback to the user. It necessarily results in false positives (failures or warnings where none are appropriate) or false negatives (no warnings when there are actual conflicts).

> or if another daemon is blocking the ports we need. The current messages when running multiple instances are cryptic.

We already capture and reflect each possible failure condition that can result from concurrent execution of similar code, and terminate: logs locked, database locked, p2p listener port in use, and server port(s) in use. Changing log messages is trivial and the necessary solution to the problem raised.

> If you want to run another instance, you need to change the blockchain path, the config files and all the ports. It's hardly another big hurdle to ask people to change or disable the PID path.

It is an *unnecessary* hurdle and ultimately dead code. Both are blocking issues IMO.

This technique not only fails to identify actual failure conditions from concurrent use, it also fails to identify them from other typical causes. The log files could be  locked by an editor or other tooling on the platform. The p2p listener can be blocked by Bitcoin Core or any other Bitcoin code, unrelated to our own. The server ports can be blocked by any application. In fact the default (Airbitz) server was recently reconfigured to run on port 80!

This technique is appropriate for the case where the service must run as a singleton on the host machine. But that is not a concern of ours. The technique is not useful in resolving the issues raised, will produce false warnings, and increases development and end-user complexity.

I understand now the problem that is being addressed, but as I see it the only way to actually resolve the problem is to improve error reporting.

RojavaCrypto:
Well it's a minor thing and not too important. Sorry this pull request has been a mess. I'm not used to managing branches in Git (it's been a while). Anyway let me know about the remaining resource_lock. If it's merged, I'll replace its usage in libbitcoin-database after.

RojavaCrypto:
"This technique not only fails to identify actual failure conditions from concurrent use, it also fails to identify them from other typical causes." - that's not true. It does the opposite. It tells you that multiple bs instances are running. It doesn't even have to be an error. It could be a warning, as opposed to non-specific messages like "blockchain db in use" or "ports already in use". But it's not important to me.

evoskuil:
No worries. We'll need to add it to the VS project before merging a dependency in database, but otherwise looks fine.

PR #429, opened 2016-08-11
Update to support 31402 to 31800 conditional serialization.

pmienk:


src/chain/header.cpp
```diff
@@ -224,7 +230,22 @@ uint64_t header::serialized_size(bool with_transaction_count) const
...
+    mutex_.lock_upgrade();
+
+    if (hash_ == nullptr)
+    {
+        mutex_.unlock_upgrade_and_lock();
+
+        if (hash_ == nullptr)
+            hash_.reset(new hash_digest(bitcoin_hash(to_data(false))));
+
+        mutex_.unlock();
```
evoskuil: This needs to be `mutex_.unlock_and_lock_upgrade();`

src/chain/transaction.cpp
```diff
@@ -263,7 +269,22 @@ std::string transaction::to_string(uint32_t flags) const
...
+    mutex_.lock_upgrade();
+
+    if (hash_ == nullptr)
+    {
+        mutex_.unlock_upgrade_and_lock();
+
+        if (hash_ == nullptr)
+            hash_.reset(new hash_digest(bitcoin_hash(to_data())));
+
+        mutex_.unlock();
```
evoskuil: Same here.

pmienk:
Thanks for the catch, corrected.

evoskuil:
I've rebased this on top of my last merged PR and added a patch, see:  #431

PR #460, opened 2016-09-06
Update libpng uri/version.

pmienk:


thecodefactory:
@pmienk Thanks for getting to this before me 👍

PR #583, opened 2016-11-11
Fix instance of potential operation access before operation cache load

thecodefactory:
Comment Typo fix

Related to https://github.com/libbitcoin/libbitcoin-explorer/pull/399

### DIFF
--- a/src/chain/script/script.cpp
+++ b/src/chain/script/script.cpp
@@ -854,7 +854,7 @@ operation::list script::to_pay_multisig_pattern(uint8_t signatures,
 script_pattern script::pattern() const
 {
     // The first operations access must be method-based to guarantee the cache.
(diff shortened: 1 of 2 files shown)


src/chain/script/script.cpp
```diff
@@ -854,7 +854,7 @@ operation::list script::to_pay_multisig_pattern(uint8_t signatures,
 script_pattern script::pattern() const
 {
     // The first operations access must be method-based to guarantee the cache.
-    if (is_null_data_pattern(operation::list()))
+    if (is_null_data_pattern(operations()))
```
thecodefactory: Right.  It was an easy bug to miss because it's syntactically correct and the comment was mis-leading ;-)

PR #590, opened 2016-11-13
Fix up statsd for VC++, fix refcounts, style.

pmienk:


evoskuil:
Thanks! Templates don't get compiled by VC++ unless they are used, so I expected there would be issues until I got to node/server.

PR #633, opened 2016-12-22
Rename output_point::validation to validation_type

RojavaCrypto:
Currently I cannot create a pointer of the libbitcoin::chain::output_point::validation* type because there is a value in that class with the same name and the compiler gets confused. This pull request renames the type with a suffix so it can be specified.

### DIFF
--- a/include/bitcoin/bitcoin/chain/output_point.hpp
+++ b/include/bitcoin/bitcoin/chain/output_point.hpp
@@ -38,7 +38,7 @@ public:
 
     // This validation data IS copied on output_point copy/move.
     // These properties facilitate block and transaction validation.
-    struct validation
+    struct validation_type
     {
         /// This is a .height sentinel.
         static const size_t not_specified;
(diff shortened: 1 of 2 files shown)


evoskuil:
This is a break in downstream repositories. It also creates an inconsistency with several other chain types. It's also a matter of style that we avoid using arbitrary prefixes and suffixes to avoid name conflicts.

The validation fields are not intended for public use. They exist only as a temporary workaround. The fact that they are mutable makes them dangerous to expose publicly. Ideally these should be exposed by derived types within libbitcoin-blockchain . That is hindered by the fact that all network messages are passed via shared pointer (an important performance optimization that prevents unnecessary copying), precluding simple derivation.

So while it is understood that the names can present a conflict in the case of addressing the field, the field should not be addressed except by class internals (effectively private) and blockchain validation (effectively friend). They would currently be friend/private (and therefore not publicly exposed) if the classes were in the same repository.

So my recommendation is to not access the field at all, and to revert the change in order to fix the build breaks.

RojavaCrypto:
Unfortunately I'm trying to port this API to the C and Python layers, and to get the unit tests working, I need to be able to access this field. I don't mind using an ugly syntax to access this field but I need a pointer to that data type:

https://github.com/libbitcoin/libbitcoin-c/blob/master/include/bitcoin/bitcoin/c/internal/chain/output_point.hpp#L30

It is used a lot in the script and transaction unit tests.

PR #635, opened 2016-12-30
Added testnet_p2kh/p2sh constants to payment_address

RojavaCrypto:


RojavaCrypto:
Actually it seems trivial enough and logical so I'll pull.

PR #636, opened 2017-01-01
Add functions for working with variable length address strings.

RojavaCrypto:
Add functions for working with variable length address strings. Getting their type. Adds these functions:

    enum class address_status_type
    {
        mainnet_p2kh,
        mainnet_p2sh,
        testnet_p2kh,
        testnet_p2sh,
        stealth,
        invalid
    };

    /// Return the type of address.
    address_status_type address_status(const std::string& address);

    /// Whether the string is a mainnet Bitcoin address.
    bool is_mainnet_address(const std::string& address);
    /// Whether the string is a testnet Bitcoin address.
    bool is_testnet_address(const std::string& address);
    /// Whether the string is a stealth address.
    bool is_stealth_address(const std::string& address);

    /// Whether the string is a valid address (status is not invalid).
    bool is_valid_address(const std::string& address);

This is added to bitcoin/wallet/address_status.hpp

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -169,6 +169,7 @@ src_libbitcoin_la_SOURCES = \
     src/utility/thread.cpp \
     src/utility/threadpool.cpp \
     src/utility/work.cpp \
+    src/wallet/address_status.cpp \
     src/wallet/bitcoin_uri.cpp \
     src/wallet/dictionary.cpp \
     src/wallet/ec_private.cpp \
@@ -562,6 +563,7 @@ include_bitcoin_bitcoin_utility_HEADERS = \
 
 include_bitcoin_bitcoin_walletdir = ${includedir}/bitcoin/bitcoin/wallet
 include_bitcoin_bitcoin_wallet_HEADERS = \
+    include/bitcoin/bitcoin/wallet/address_status.hpp \
     include/bitcoin/bitcoin/wallet/bitcoin_uri.hpp \
     include/bitcoin/bitcoin/wallet/dictionary.hpp \
     include/bitcoin/bitcoin/wallet/ec_private.hpp \
--- a/include/bitcoin/bitcoin/wallet/address_status.hpp
+++ b/include/bitcoin/bitcoin/wallet/address_status.hpp
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) 2011-2015 libbitcoin developers (see AUTHORS)
+ *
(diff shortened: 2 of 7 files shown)


evoskuil:
Ugh, I did not intend to hit the close button...

evoskuil:
I'm inclined to recommend against this change because it is undergeneralized for libbitcoin::wallet.

As a rule wallet supports all altchains equally, to the extent they follow Bitcoin conventions and deviate only by "prefix" value. There are symbolic identifiers for the two most commonly-used prefixes, but these are in no way limiting. The change purports to enumerate valid address types based on a limited set of prefixes. To properly generalize such a facility one needs to parse the string and return the prefix, since the prefix *is* the enumeration. 

A similar problem exists with the attempt to generalize parsing across payment_address and stealth_address types. The reason neither of these types is called wallet::address is that name is reserved for this generalization. We specifically work to avoid collections of free functions. The way that would be accomplished in this case is to implement the generalized wallet::address class.

The reason I have not implemented wallet::address is that it is a good amount of work for a weak generalization and it is not at all certain that stealth will survive.

The addition of wallet class is_valid() methods was previously discussed and as I said, I am not against it. But the change must be consistently applied to all wallet types, and it must be added to the wallet test matrix in each type. At that point the then-redundant Boolean cast operators should be removed and all affected code updated. Just adding the two methods is an (unnecessary) inconsistency that creates more work for someone else.

I would recommend retaining these changes as application utilities until such time that bc::wallet::address is implemented (or bc::wallet::stealth_address is deprecated).

RojavaCrypto:
OK! As long as you understand the motivation behind these methods- primarily validation of user entered Bitcoin addresses into the send field of a Bitcoin wallet.

evoskuil:
I do, and I can see a generalization of payment_address and stealth_address as useful in some scenarios. I've contemplated doing this but the generalization is very weak. It's practically limited to just validation of the text input. After that point there is little if any commonality.

That check is very simple using current classes. I would always parse using `payment_address` and if the parse fails fall back to `stealth_address`. Then a simple test against the version values supported by the wallet allows you to provide proper detailed feedback to the user. In all this is just a few lines of code.

Note that there are other artifacts that a wallet will want to import. There are several forms of private and public keys, including encrypted and non-encrypted, and various encodings, including mnemonic. One could argue that a single text input field would make importing of these values easier. IMO this is the job of the application, using the primitives provided by libbitcoin and the method described above.

PR #652, opened 2017-01-10
Rem dead code, add missing_previous_transactions, change missing_prev…

evoskuil:
…ious_outputs.


RojavaCrypto:
Is this output_point::list or point::list because the former is ambiguous. The typedef in the class is actually inherited as std::vector<point>.

PR #659, opened 2017-01-15
Simplified Linux install instructions.

RojavaCrypto:


evoskuil:
Note that we have similar build instructions in [bs](https://github.com/libbitcoin/libbitcoin-server/blob/master/README.md) and [bx](https://github.com/libbitcoin/libbitcoin-explorer/blob/master/README.md) that we need to keep in sync. In all three locations there are numerous examples that start with:
```sh
$ ./install.sh ...
```
It's nice to remove the `chmod` step but I think it would be best to retain the install.sh downloaded file in order to not break the flow for those subsequent examples. Would this be ok? If so could you instead devise a simple syntax to eliminate the chmod?

Thanks!

evoskuil:
This ends up just dropping the chmod step, which would be a break. I was hoping for a switch on wget to obviate the step, but i don't see anything that would be less complicated than just leaving the original.

PR #668, opened 2017-01-20
Code review and updates for stealth utilities.

evoskuil:
The new stealth utilities have been split into `stealth_spender` and `stealth_receiver` (as per the file-per-class rule). The `stealth_sender` utility has been modified to move mutators into construction, resulting in a simpler const class with no material loss of utility.

The existing stealth utility `create_stealth_data` has been renamed to `create_stealth_script` and updated to return the full stealth null_data script as opposed to just its contained data. It has also been overloaded to allow for parameterization of an existing ephemeral key pair, which is necessary for HD wallets.

The updated `create_stealth_script` utility has been applied in the `stealth_sender` function so that it does not reproduce this functionality.

The new utilities are marked "experimental" as they have four material deficiencies:

1. All error handling is performed by assertion. This is sufficient for test and example code but not in production. Assertions must be limited to testing developer assumptions - specifically the belief that the assert *cannot* be triggered if the code is implemented as intended. Any assertion that may be triggered by parameterization of a public API introduces an instability that we try hard to avoid.

2. There are two cases of embedded seed generation that pertain to security and privacy. Libbitcoin always parameterizes seeds so that the developer is not forced into a dependency on a weak RNG.

3. The stealth implementation does not support a prefix filter. This requires that the user scan all potential stealth transactions on the blockchain.

4. The stealth implementation accepts a stealth address, which includes multiple spend keys, but internally uses only the first spend key.

Given the updated `create_stealth_script` it appears that the simplification remaining from these two const wrapper classes arises entirely from the above issues. The developer is freed from error handling, filtering, RNG production and multi-spender support. However with the possible exception of the last, these are not things that we want to hide from the developer. Reintroducing these will eliminate any simplification offered by these utilities. It's for this reason I recommend removing them form the library before v3 release unless we have a better resolution by then.


RojavaCrypto:
Perfect. About being limited to the first spend key, I assume we will need a separate class to handle stealth multisig since the API calling conventions are sufficiently different.

RojavaCrypto:
The wrappers might look cosmetic, but you have to compare a bunch of calls like:
```
uncover_stealth(out, scan, ephemeral);
uncover_stealth(out, spend, ephemeral_private);
create_stealth_script(script, filter, seed);
```
Which doesn't improve the readability of code to express what is being done. This code creates separation of concerns (receiver and sender abstractions), but if you want to remove it, that's also fine. We can put this in a higher level library too.

evoskuil:
The objection is not that the wrappers are cosmetic but that there is no error handling, seeds are internally-generated, and there is no ability to provide a stealth filter. Even if we accept the restricted parameterization (or elevate that to another class for handling multisig) each of these other issues remains significant. I'll see if I can resolve them when I get a chance.

RojavaCrypto:
1) it's an inevitable consequence that lower level code gets folded into higher level code where errors simply become simple pass/fail events. 2) seeds can also be optionally passed. 3) that's correct.

evoskuil:
I disagree entirely with the idea of abandoning error handling at higher level code. That is just guaranteeing instability. There is no place in the entire libbitcoin codebase that avoids error handling.

Stealth without filters is a plan to scan the entire blockchain worth of potential stealth tx metadata. Do you mean to suggest that stealth filters should be avoided?

evoskuil:
See https://github.com/libbitcoin/libbitcoin/pull/672

PR #683, opened 2017-02-02
Adds French and Italian mnemonic word lists (BIP39)

thecodefactory:


test/wallet/mnemonic.cpp
```diff
@@ -131,6 +131,70 @@ BOOST_AUTO_TEST_CASE(mnemonic__dictionary__en_es__no_intersection)
...
+    {
+        std::string test(it);
+        const auto iter = std::find(french.begin(), french.end(), test);
+        if (iter != std::end(french))
+            intersection++;
+    }
+
+    BOOST_REQUIRE_EQUAL(intersection, 0u);
+}
+
```
evoskuil: We should probably use [std:set_intersection](http://en.cppreference.com/w/cpp/algorithm/set_intersection) for these.

PR #795, opened 2017-07-21
Implement BIP68, BIP112 and BIP113 (CSV).

evoskuil:
This also expands script textual encoding to include signed 64 bit decimal values as data token, non-minimal data encoding, quoted string as data token, allowance for oversize data tokens, and empty scripts. Apart from our intentional non-support of multi-token data encodings, and use of [] vs. 0x for hex data encoding, these changes make libbitcoin script a superset of satoshi client script.


evoskuil:
Oops, wrong branch.

PR #806, opened 2017-09-15
Allow version information when converting from script

thecodefactory:
Allow version information when converting from script to payment address


thecodefactory:
Given an output, this is an alternative that allows output.address(version)) rather than calling something like payment_address::from_script(output.script(), version) and seems to better match given that the address() and addresses() methods of the output were provided already (just minus some flexibility).

PR #817, opened 2017-10-21
Update documentation for ubuntu 16.04

### LINKED ISSUE #805: Undefined reference to boost
I'm trying to run a simple program that includes libbitcoin and I get this error when compiling:

```
g++-4.8 -std=c++11 -I /usr/include/boost/include -o hello hello.cpp
/tmp/cczhQ2e3.o: In function `__static_initialization_and_destruction_0(int, int)':
hello.cpp:(.text+0x5f): undefined reference to `boost::system::generic_category()'
hello.cpp:(.text+0x6b): undefined reference to `boost::system::generic_category()'
hello.cpp:(.text+0x77): undefined reference to `boost::system::system_category()'
hello.cpp:(.text+0xb5): undefined reference to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)'
hello.cpp:(.text+0x104): undefined reference to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)'
hello.cpp:(.text+0x153): undefined reference to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)'
hello.cpp:(.text+0x1a2): undefined reference to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)'
hello.cpp:(.text+0x1f1): undefined reference to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)'
/tmp/cczhQ2e3.o:hello.cpp:(.text+0x240): more undefined references to `libbitcoin::config::checkpoint::checkpoint(std::string const&, unsigned long)' follow
/tmp/cczhQ2e3.o: In function `__static_initialization_and_destruction_0(int, int)':
hello.cpp:(.text+0x318): undefined reference to `libbitcoin::machine::number::negative_mask'
/tmp/cczhQ2e3.o: In function `boost::asio::error::get_system_category()':
hello.cpp:(.text._ZN5boost4asio5error19get_system_categoryEv[_ZN5boost4asio5error19get_system_categoryEv]+0x5): undefined reference to `boost::system::system_category()'
collect2: error: ld returned 1 exit status
```
I use Ubuntu 16.04.3, and followed install instructions in the README of https://github.com/libbitcoin/libbitcoin. 

I installed libbitcoin using `sudo ./install.sh --build-boost --prefix=/usr/include/boost` but boost path isn't apparently recognised when compiling hello.cpp.

The program is:
```
#include <iostream>
#include <bitcoin/bitcoin.hpp>

int main() {
  std::cout << "Hello Bitcoin";
}
```
Am I missing something ?

Thanks in advance

michel-foucault:
resolved #805

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,8 @@ $ sudo update-alternatives --install /usr/bin/g++ g++ /usr/bin/g++-4.8 50
 $ sudo update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-4.8 50
 $ sudo update-alternatives --install /usr/bin/gcov gcov /usr/bin/gcov-4.8 50
 ```
+_Note: For ubuntu 16.04 use default g++ compiler (g++-5), avoiding boost incompatibility_
+
 Next install the [build system](http://wikipedia.org/wiki/GNU_build_system) (Automake minimum 1.14) and git:
 ```sh
 $ sudo apt-get install build-essential autoconf automake libtool pkg-config git


evoskuil:
@michel-foucault  - thanks for this information and update, however we generally prefer to resolve the root issue in the installer so that we don't end up with a variety of build workarounds. I believe that @pmienk is giving this a closer look.

michel-foucault:
I'm ok with that,
I wondered if the issue was only using g++-4.8 in command line with missing dependencies.
I had double-checked and it seems that even with the full dependencies in the command line, using boost (1.58.0) from apt-get on linux 16.04, cause the issue.
It should work with --with-boost option (didn't check).

But in the end the Install.sh is not buggy, it is an development environment issue (downgrading g++, using g++ directly without Makefile, ...).

evoskuil:
Thanks again @michel-foucault. I'm closing this but it will remain referenced from the open boost issue.

PR #828, opened 2017-11-02
Regenerate installer.

evoskuil:


evoskuil:
See: https://github.com/libbitcoin/libbitcoin/pull/829

PR #838, opened 2017-11-28
Regtest network support

lastcanal:
This PR adds everything required to run libbitcoin on the regtest network as described by @evoskuil in the issue https://github.com/libbitcoin/libbitcoin-server/issues/404. This includes adding the encoded regtestnet genesis block, updated proof of work limit, and work required calculation. 

All of these changed are protected by the `WITH_REGTEST` flag, except `encoded_regtestnet_genesis_block` and `genesis_regtestnet ` which are namespaced.

This PR doesn't implement any makefile changes. Adding `#define WITH_REGTEST` near the top of `include/bitcoin/bitcoin/constants.hpp` works as a temporary workaround.

### DIFF
--- a/include/bitcoin/bitcoin/chain/block.hpp
+++ b/include/bitcoin/bitcoin/chain/block.hpp
@@ -138,6 +138,7 @@ public:
 
     static block genesis_mainnet();
     static block genesis_testnet();
+    static block genesis_regtestnet();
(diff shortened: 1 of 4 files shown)


evoskuil:
Thanks Ty! I'll smooth out the configurability of this so that it does not require conditional compilation in any lib.

PR #843, opened 2017-12-03
cmake: Add cmake support

aschuess:
These changes allow the user to generate makefiles and VS sln files via CMake. 

It currently only builds static library.
This enables support for visual studio 2017.

Enables ease of integration with other cmake projects.


evoskuil:
This is great @aschuess! However... we cannot merge this directly into libbitcoin. All build artifacts are generated using the [libbitcoin-build](https://github.com/libbitcoin/libbitcoin-build) system, and we require all builds to work across all repositories and for CI validation of builds. The maintenance load is unbearably high and results over time are unreliable if not generated and tested.

GSL can be a bit of a challenge. @pmienk Has recently completed the work to directly generate VS2013 artifacts, which is is planning to to apply across multiple versions of VS (i.e. 2013/2015/2017). At that point it may be worth adding build templates for CMake based on your work.

aschuess:
Sounds good. Agreed maintaining two build systems is not a great idea; more wanted you to be aware it's available if someone needs cmake support. 

The ease of adding it is a testament to the good organization or the project. Kudos @evoskuil and team.

evoskuil:
Thanks for the feedback @aschuess. We currently generate support 2 build systems (Visual Studio and Autotools) but expect to do the same for CMake and XCode as code generation makes this fairly straightforward. The main limitation is having a template to work from, which you have now provided for CMake :).

evoskuil:
Moved to open issue in libbitcoin-build. Thanks for submitting, we should get to this in the not-too-distant future, across all repositories!

PR #846, opened 2017-12-05
Initial add of electrum 2.x mnemonic support

thecodefactory:


src/wallet/electrum.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return mnemonic;
+}
+
+static cpp_int mnemonic_decode(const word_list& mnemonic,
+    const dictionary& lexicon)
+{
+    cpp_int entropy = 0;
+    const auto dictionary_length = lexicon.size();
+
+    for(auto word = mnemonic.rbegin(); word != mnemonic.rend(); word++)
```
evoskuil: Style: use boost's `reverse(...)` iteration template here.

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,234 @@
...
+    return false;
+}
+
+#endif
+
+// This operation does not require normalization support
+long_hash decode_mnemonic(const word_list& mnemonic)
+{
+    const auto sentence = join(mnemonic);
+    const auto salt = passphrase_prefix;
```
evoskuil: Use reference here (don't copy).

test/wallet/electrum.cpp
```diff
@@ -0,0 +1,122 @@
...
+// These test cases were generated using electrum code and verified to
+// match output from:
+//
+// https://github.com/spesmilo/electrum/blob/master/lib/mnemonic.py
+
+BOOST_AUTO_TEST_CASE(electrum__decode_mnemonic__no_passphrase_1)
+{
+    const auto mnemonic = word_list{"foo", "bar", "baz"};
+    const auto seed = electrum::decode_mnemonic(mnemonic);
+    BOOST_REQUIRE_EQUAL(encode_base16(seed),
```
evoskuil: Style: don't wrap code lines of test cases (braced data initializers okay).

test/wallet/electrum.cpp
```diff
@@ -0,0 +1,122 @@
...
+BOOST_AUTO_TEST_SUITE(electrum_tests)
+
+// These test cases were generated using electrum code and verified to
+// match output from:
+//
+// https://github.com/spesmilo/electrum/blob/master/lib/mnemonic.py
+
+BOOST_AUTO_TEST_CASE(electrum__decode_mnemonic__no_passphrase_1)
+{
+    const auto mnemonic = word_list{"foo", "bar", "baz"};
```
evoskuil: Style:
```cpp
static const auto word_list mnemonic{ { "foo", "bar", "baz"} };
```

test/wallet/electrum.cpp
```diff
@@ -0,0 +1,122 @@
...
+    BOOST_REQUIRE_EQUAL(encode_base16(seed),
+        "741b72fd15effece6bfe5a26a52184f66811bd2be363190e07a42cca442b1a5bb22b3ad0eb338197287e6d314866c7fba863ac65d3f156087a5052ebc7157fce");
+}
+
+BOOST_AUTO_TEST_CASE(electrum__create_mnemonic__dictionary)
+{
+    data_chunk entropy;
+    decode_base16(entropy, "05E669B4270F4E25BCE6FC3736170D423C");
+    const auto mnemonic = electrum::create_mnemonic(entropy, language::en);
+    BOOST_REQUIRE(mnemonic.size() > 0);
```
evoskuil: Use `BOOST_REQUIRE(!mnemonic.empty());` here.

include/bitcoin/bitcoin/wallet/electrum.hpp
```diff
@@ -0,0 +1,110 @@
...
+
+/**
+ * Valid seed types.
+ */
+enum class seed_type : uint8_t
+{
+    standard = 1,
+    segwit = 2,
+    two_factor_authentication = 3
+};
```
evoskuil: Avoid use of "type" in types. `bc::wallet::electrum::seed` should be sufficient. Avoid the "segwith" acronym, instead use "witness".

thecodefactory:
Updated to incorporate feedback.

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+{
+    word_list mnemonic;
+    const auto dictionary_length = lexicon.size();
+
+    while(entropy)
+    {
+        const cpp_int index = entropy % dictionary_length;
+        entropy /= dictionary_length;
+        mnemonic.push_back(lexicon[static_cast<size_t>(index)]);
+    }
```
thecodefactory: The cpp_int modulo operation returns a cpp_int, which needs to be casted when used as an index.  It doesn't allow me to use a size_t for the assignment.

evoskuil: What's driving the need for cpp_int? Also note that boost uses cpp_int as a base type for more specific numeric types.

thecodefactory: We need arbitrary precision integers for this to work.  cpp_int is a base type for many kinds of fixed precision numeric types, but not as many kinds of arbitrary precision integers.

If we use a fixed precision numeric type, we cannot match the output produced from the electrum python reference implemention when used with a bigger number. 

Also, I don't know of a way to use unsigned arbitrary precision cpp_int types in boost, as specified by their documentation.

Makefile.am
```diff
@@ -182,6 +182,8 @@ src_libbitcoin_la_SOURCES = \
     src/wallet/ek_private.cpp \
     src/wallet/ek_public.cpp \
     src/wallet/ek_token.cpp \
+    src/wallet/electrum.cpp \
+    src/wallet/electrum_dictionary.cpp \
```
evoskuil: There should be an `electrum_dictionary.hpp` - see outdated comments.

include/bitcoin/bitcoin/wallet/dictionary.hpp
```diff
@@ -60,6 +67,12 @@ namespace language
 
     // All built-in languages:
     extern const dictionary_list all;
+
+    namespace electrum
+    {
+        // electrum v1 dictionary required for v2 compatibility
+        extern const v1_dictionary en_v1;
```
evoskuil: Just a nit, but can we be consistent with the version as a suffix?
```cpp
extern const dictionary_v1 en_v1;
```

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+    const auto dictionary_length = lexicon.size();
+    const auto seed_size = sizeof(uint32_t) * mnemonic.size() / 3;
+
+    data_chunk seed;
+    seed.reserve(seed_size);
+    data_sink ostream(seed);
+    ostream_writer sink(ostream);
+
+    auto all_found = false;
+    for(size_t i = 0; i < mnemonic.size() / 3; i += 3)
```
evoskuil: Whitespace for non-function semantics.

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+    {
+        const auto index_1 = find_position(lexicon, mnemonic[i]);
+        const auto index_2 = find_position(lexicon, mnemonic[i+1]);
+        const auto index_3 = find_position(lexicon, mnemonic[i+2]);
+
+        all_found = ((index_1 != -1) && (index_2 != -1) && (index_3 != -1));
+        if (!all_found)
+            break;
```
evoskuil: Recommend simply `return{}` here and elimination of the other code pertaining to `all_found`.

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+        mnemonic.size() == mnemonic_size_v1_1 ||
+        mnemonic.size() == mnemonic_size_v1_2);
+}
+
+static bool is_new_seed(const word_list& mnemonic, const data_slice& prefix)
+{
+    const auto seed = hmac_sha512_hash(
+        to_chunk(to_normal_nfkd_form(join(mnemonic))), hmac_data);
+
+    return (std::memcmp(seed.data(), prefix.data(), prefix.size()) == 0);
```
evoskuil: Avoid memcmp as it can be unsafe depending on circumstance. It's better to use std::equal where possible (as in this case).

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+}
+
+static word_list mnemonic_encode(cpp_int entropy, const dictionary& lexicon)
+{
+    word_list mnemonic;
+    const auto dictionary_length = lexicon.size();
+
+    while(entropy)
```
evoskuil: Whitespace.

src/wallet/electrum.cpp
evoskuil: Okay, makes sense. And the case is of course protected by the % operation, so looks good.

src/wallet/electrum.cpp
```diff
@@ -0,0 +1,237 @@
...
+    return mnemonic;
+}
+
+static cpp_int mnemonic_decode(const word_list& mnemonic,
+    const dictionary& lexicon)
+{
+    cpp_int entropy = 0;
+    const auto dictionary_length = lexicon.size();
+
+    for(auto& word : boost::adaptors::reverse(mnemonic))
```
evoskuil: Use `const auto&`

test/wallet/electrum.cpp
```diff
@@ -0,0 +1,110 @@
...
+    static const word_list mnemonic{ { "foo", "bar", "baz"} };
+    const auto seed = electrum::decode_mnemonic(mnemonic);
+    BOOST_REQUIRE_EQUAL(encode_base16(seed), "c4033901dd61ba26cfd0a1cf1ceb4b347606635aa3cb951eb6e819d58beedc04dd400a2e600d783c83c75879d6538abeecc7bb1b292b2a4d775d348d5d686427");
+}
+
+BOOST_AUTO_TEST_CASE(electrum__decode_mnemonic__no_passphrase_2)
+{
+    static const word_list mnemonic{ { "giggle", "crush", "argue", "inflict",
+        "wear", "defy", "combine", "evolve", "tiger", "spatial", "crumble", "fury" } };
+    const auto seed = electrum::decode_mnemonic(mnemonic);
```
evoskuil: I was probably ambiguous in the last style recommendation on this point. The "wrap" allowed in this case is for balanced bracing... basically for more readable declaration of complex structures, maybe not so helpful here:
```cpp
static const word_list mnemonic
{
    {
        "giggle", "crush", "argue", "inflict", "wear", "defy", "combine", "evolve", "tiger", "spatial", "crumble", "fury"
    }
};
```

thecodefactory:
Updated to incorporate additional feedback.

include/bitcoin/bitcoin/wallet/electrum.hpp
thecodefactory: Addressing this now

PR #852, opened 2017-12-08
Regenerate artifacts.

pmienk:


evoskuil:
Looks like this did not resolve the issue: https://travis-ci.org/libbitcoin/libbitcoin/jobs/313792706

PR #854, opened 2017-12-09
Wrap only required passphrase usage with the WITH_ICU conditional

thecodefactory:


thecodefactory:
This should fix the build issue in libbitcoin/libbitcoin-explorer#516

PR #890, opened 2018-02-04
Added base32

marcoferiencik:
The new files add bech32 encoding/decoding for data chunks according to bip173.

### DIFF
--- a/include/bitcoin/bitcoin/formats/base_32.hpp
+++ b/include/bitcoin/bitcoin/formats/base_32.hpp
@@ -1,0 +1,46 @@
+/**
+ * Copyright (c) 2011-2018 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_BASE_32_HPP
+#define LIBBITCOIN_BASE_32_HPP
+
+#include <string>
+#include <bitcoin/bitcoin/define.hpp>
+#include <bitcoin/bitcoin/utility/data.hpp>
+
+namespace libbitcoin {
+
+/**
+ * Encode data as base32.
+ * @return the base32 encoded string.
+ */
+BC_API std::string encode_base32(const std::string& hr_part, const data_chunk& data_part);
+
+/**
+ * Decode base32 data.
+ * @param in base32 string to decode.
+ * @param hr_part The decoded human-readable part.
+ * @param values The decoded values.
+ * @return false if the input is not a valid base32 encoded string.
+ */
+BC_API bool decode_base32(std::string& hr_part, data_chunk& values, const std::string& in);
+
+} // namespace libbitcoin
+
+#endif
+
--- a/src/formats/base_32.cpp
+++ b/src/formats/base_32.cpp
@@ -1,0 +1,174 @@
+/**
+ * Copyright (c) 2011-2018 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <bitcoin/bitcoin/formats/base_32.hpp>
+
+#include <cstdint>
+
+namespace libbitcoin {
+
+const static char lut[] = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+const int8_t rev_lut[128] = {
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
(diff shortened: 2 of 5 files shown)


test/formats/base_32.cpp
```diff
@@ -0,0 +1,65 @@
...
+
+using namespace bc;
+
+BOOST_AUTO_TEST_SUITE(base_32_tests)
+
+bool decode_test(const std::string& in)
+{
+    std::string hr_part;
+    data_chunk values;
+    bool valid = libbitcoin::decode_base32(in, hr_part, values);
```
toxeus: This should be 
```cpp
return libbitcoin::decode_base32(in, hr_part, values);
```
That's why the tests fail.

test/formats/base_32.cpp
```diff
@@ -0,0 +1,65 @@
...
+    BOOST_REQUIRE(false == decode_test("\x80""1eym55h"));
+    BOOST_REQUIRE(false == decode_test("an84characterslonghumanreadablepartthatcontainsthenumber1andtheexcludedcharactersbio1569pvx"));
+    BOOST_REQUIRE(false == decode_test("pzry9x0s0muk"));
+    BOOST_REQUIRE(false == decode_test("1pzry9x0s0muk"));
+    BOOST_REQUIRE(false == decode_test("x1b4n0q5v"));
+    BOOST_REQUIRE(false == decode_test("li1dgmt3"));
+    BOOST_REQUIRE(false == decode_test("de1lg7wt\xFF"));
+    BOOST_REQUIRE(false == decode_test("A1G7SGD8"));
+    BOOST_REQUIRE(false == decode_test("10a06t8"));
+    BOOST_REQUIRE(false == decode_test("1qzzfhee"));
```
toxeus: Usually, `!` is used instead of `false ==`.

test/formats/base_32.cpp
```diff
@@ -0,0 +1,65 @@
...
+BOOST_AUTO_TEST_CASE(base32_decode_test)
+{
+    // Valid
+    BOOST_REQUIRE(decode_test("a12uel5l"));
+    BOOST_REQUIRE(decode_test("A12UEL5L"));
+    BOOST_REQUIRE(decode_test("an83characterlonghumanreadablepartthatcontainsthenumber1andtheexcludedcharactersbio1tt5tgs"));
+    BOOST_REQUIRE(decode_test("abcdef1qpzry9x8gf2tvdw0s3jn54khce6mua7lmqqqxw"));
+    BOOST_REQUIRE(decode_test("11qqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqqc8247j"));
+    BOOST_REQUIRE(decode_test("split1checkupstagehandshakeupstreamerranterredcaperred2y9e3w"));
+    BOOST_REQUIRE(decode_test("?1ezyfcl"));
```
toxeus: I'd prefer if you'd `BOOST_REQUIRE_EQUAL` for the decoded value similiar to what you did in the tests for `encode_base32()` instead of just checking for validity.

test/formats/base_32.cpp
```diff
@@ -0,0 +1,65 @@
...
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <boost/test/unit_test.hpp>
+#include <bitcoin/bitcoin.hpp>
+
+using namespace bc;
```
toxeus: Since you're using the `bc` namespace here you can drop the namespace specifier from `decode_base32()` and `encode_base32()`.

src/formats/base_32.cpp
```diff
@@ -0,0 +1,187 @@
...
+
+inline char toLowercase(const char& c)
+{
+    return c + ('a' - 'A');
+}
+
+bool decode_base32(const std::string& in, std::string& hr_part, data_chunk& values)
+{
+    // Basic input check
+    if (in.length() < base32_min_length || in.length() > base32_max_length)
```
toxeus: Here, the check for the min length is redundant to the `pos_separator` checks on line 132. A string meeting the minimum length requirement still can be invalid.

src/formats/base_32.cpp
```diff
@@ -0,0 +1,187 @@
...
+    {
+        return false;
+    }
+    
+    // Find separator and check valid hr_part and data_part sizes
+    size_t pos_separator = in.rfind(separator);
+    if (pos_separator == std::string::npos 
+        || pos_separator < hr_part_min_length
+        || pos_separator > hr_part_max_length
```
toxeus: The check on this line is redundant because you're already checking for the input string not exceeding the maximum length on line 125 and checking for both parts having the minimum length required.

src/formats/base_32.cpp
```diff
@@ -0,0 +1,187 @@
...
+    // Convert data part to values
+    const size_t data_part_length = encoded.size() - (pos_separator+1);
+    data_chunk data_part_values;
+    data_part_values.reserve(data_part_length);
+    for (auto i = pos_separator + 1; i < encoded.size(); ++i)
+    {
+        if (rev_lut[encoded[i]] == -1)
+            return false;
+        data_part_values.push_back(rev_lut[encoded[i]]);
+    }
```
toxeus:
```cpp
for (const unsigned char c : encoded.substr(pos_separator + 1, encoded.npos))
{
    if (rev_lut[c] == -1)
        return false;
    data_part_values.push_back(rev_lut[c]);
}
```
Otherwise, you'd be indexing `rev_lut` with something of type `char`. Also, I find it more readable this way :wink:

include/bitcoin/bitcoin/formats/base_32.hpp
```diff
@@ -0,0 +1,46 @@
...
+BC_API std::string encode_base32(const std::string& hr_part, const data_chunk& data_part);
+
+/**
+ * Decode base32 data.
+ * @param in base32 string to decode.
+ * @param hr_part The decoded human-readable part.
+ * @param values The decoded values.
+ * @return false if the input is not a valid base32 encoded string.
+ */
+BC_API bool decode_base32(const std::string& in, std::string& hr_part, data_chunk& values);
```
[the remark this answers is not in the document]
evoskuil: Agree, we always set out parameters first. This prevents the consistency problem that arises later when adding an optional parameter (which must be last).

src/formats/base_32.cpp
```diff
@@ -0,0 +1,187 @@
...
+const static size_t base32_min_length = hr_part_min_length + 1 + data_part_min_length;
+const static size_t base32_max_length = 90;
+const static size_t data_part_max_length = base32_max_length - hr_part_min_length - 1;
+
+uint32_t polymod_base32(const data_chunk& values)
+{
+    static const std::vector<uint32_t> gen = 
+        {0x3b6a57b2, 0x26508e6d, 0x1ea119fa, 0x3d4233dd, 0x2a1462b3};
+    uint32_t chk = 1;
+    for (auto v : values)
```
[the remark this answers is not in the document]
evoskuil: Agree, but should be:
```cpp
for (const auto& value: values)
```

evoskuil:
Thanks @marcoferiencik! Agree with all comments from @toxeus.

I'm traveling almost the whole month, so my responses are a little slow. After reviewing the commits I notice a good number of style issues. I suggest a review of both:

https://github.com/libbitcoin/libbitcoin/wiki/Style-Guide
https://github.com/libbitcoin/libbitcoin/wiki/Test-Guide

Otherwise I would suggest managing the "human readable part" (an awful term, should be "prefix") and "data part" as a tuple. That way you and up with singular input/output parameters and better readability (value.prefix, value.data).

src/formats/base_32.cpp
marcoferiencik: Yes this is true. I didn't write explicitly, but this first max length check assures that the method returns fast in case of a very large input string whilst find() needs to run over the whole string - worst case. So I suggest to keep this check.

marcoferiencik:
Thanks for the constructive feedback! I am incorporating the comments and will also provide tests for combined encoding and decoding.

src/formats/base_32.cpp
toxeus: @marcoferiencik I agree that the max length check makes sense. I was talking about the min length check 😉

marcoferiencik: Oh, I misread, of course this makes sense. Thanks!

PR #891, opened 2018-02-09
undefined behavior fixed

fpelliccioni:
Error not detected by the unit tests

### DIFF
--- a/src/utility/png.cpp
+++ b/src/utility/png.cpp
@@ -112,8 +112,7 @@ bool png::write_png(std::istream& in, uint32_t size, uint32_t dots_per_inch,
         if (info_ptr == nullptr)
             return false;
 
-        png_color raw_palette;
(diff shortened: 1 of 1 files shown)


thecodefactory:
@fpelliccioni This looks correct.  It doesn't look like a packed struct and is an odd size, so it's likely padded for alignment.  The pointer writes probably succeed most of the time when they shouldn't.  The test would not detect this in that case.

fpelliccioni:
Please see the complete file. It is trying to access to an non existing
array element.

thecodefactory:
Right.  I was saying that your fix looks correct.

thecodefactory:
Not sure why the last travis build is hanging, but merging this.

thecodefactory:
Note: This is probably related to libbitcoin#715

evoskuil:
Nice, thanks!

PR #894, opened 2018-02-16
Move some common explorer config classes into system

thecodefactory:
Move some common explorer config classes and prop_tree methods into core as property_tree


thecodefactory:
Related to libbitcoin/libbitcoin-explorer#528

PR #899, opened 2018-02-20
Fix up electrum test case.

evoskuil:


evoskuil:
@thecodefactory I'm not sure why yet, but there is is some sort of non-deterministic behavior in at least one electrum test case.

thecodefactory:
@evoskuil I can't isolate it from the log exactly, but I'm fairly certain the electrum test isn't getting run in the travis build that passes.  The number of tests is smaller than it should be, but that doesn't prove the electrum ones in particular are omitted.  Is there a chance WITH_ICU isn't being defined (which would skip some electrum and mnemonic tests)?  That could explain it, but it could be something else as well.  The failed builds correctly fail based on the incorrect word list (except that last one that stalls of course).

evoskuil:
That is possible, since the Travis builds have varied parameterization.

thecodefactory:
Agreed.  I think without this PR, all will build properly again (except that last one).  What was the issue that required this change?  I've found no configuration locally where this works.

evoskuil:
It was required by the VC++ build.

PR #913, opened 2018-03-05
Random re-factor

thecodefactory:
See https://github.com/libbitcoin/libbitcoin-network/pull/193


thecodefactory:
Work in progress, apparently.  Odd that it works locally, given the errors.

thecodefactory:
Updated.  In local trees, I'm able to use random::get_twister when a generator is needed, which was the original intent.  Since there are multiple ways to re-factor this and this is just one, feedback is appreciated.

The downside is that this work is for supporting https://github.com/libbitcoin/libbitcoin-network/pull/193, but it may cause v3 compatibility issues.

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -19,40 +19,52 @@
...
- */
-BC_API uint64_t pseudo_random();
+class random
+{
+  public:
+    /**
+     * Returns a thread-safe static random generator instance.
+     * @return  The generator.
+     */
+    static std::mt19937& get_twister();
```
evoskuil: The idea of making random a class is to keep this and any other implementation details private while still being able to exposing them to templatized public members.

thecodefactory: What I like about how it is now is that the methods are exposed and can be templatized and any caller can now just replace *pseudo_random* with *random::pseudo_random* instead of creating a local instance and then calling that method on it.

evoskuil: Since they were functions previously there was no instantiation. The only difference is in the predication of the class name. The v3 implementation can retain compatibility (eliminating the class name predication requirement) with a `using` declaration or a wrapper function.

evoskuil: It is also worth reviewing the method names, given that they are collected in the scope of `random`. The word `random` in each function name is a redundancy that should be removed.

thecodefactory:
For the names, what about something like:

```
pseudo_random() -> next()
pseudo_random(begin, end) -> range_next(begin, end)
pseudo_random_fill(container) -> fill(container)
pseudo_randomize(expiration, ratio) -> range_time_next(expiration, ratio)
```

evoskuil:
A couple simplifications:
```
pseudo_random() -> next()
pseudo_random(begin, end) -> next(begin, end)
pseudo_random_fill(container) -> fill(container)
pseudo_randomize(expiration, ratio) -> duration(expiration, ratio)
```

thecodefactory:
Wrapper functions instead of *using* is a bit ugly since it doubles up the functions, but it would allow us to cleanly deprecate the existing ones and keep compatibility for v3 (and then drop them in v4 of course).

evoskuil:
You could comment the same deprecation against the using declarations.

evoskuil:
Also, the reason for using “pseudo” in school name was to emphasize the potentially-unsafe lack of true randomness. I think we should preserve that in the class name.

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -50,10 +57,52 @@ BC_API void pseudo_random_fill(data_chunk& chunk);
...
+     * @return  The 64 bit number.
+     */
+    static BC_API uint64_t next(uint64_t begin, uint64_t end);
+
+    /**
+     * Fill a buffer with randomness using the default random engine.
+     * @param[in]  out  The datatype to fill with randomness.
+     */
+    template<class Container>
+    static BC_API void fill(Container& out);
```
evoskuil: Templates should be listed before statics.

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -50,10 +57,52 @@ BC_API void pseudo_random_fill(data_chunk& chunk);
...
 
+
+class pseudo_random
+{
+  public:
+    /**
+     * Returns a thread-safe static random generator instance.
+     * @return  The generator.
+     */
+    static std::mt19937& get_twister();
```
evoskuil: This should be private.

src/utility/random.cpp
```diff
@@ -62,30 +77,39 @@ static std::mt19937& get_twister()
...
-uint64_t pseudo_random(uint64_t begin, uint64_t end)
+uint64_t pseudo_random::next(uint64_t begin, uint64_t end)
 {
     std::uniform_int_distribution<uint64_t> distribution(begin, end);
     return distribution(get_twister());
 }
 
-void pseudo_random_fill(data_chunk& chunk)
+template<class Container>
+void pseudo_random::fill(Container& out)
```
evoskuil: The template implementation cannot go in the .cpp file.

src/utility/random.cpp
```diff
@@ -62,30 +77,39 @@ static std::mt19937& get_twister()
...
 {
     // uniform_int_distribution is undefined for sizes < 16 bits.
     std::uniform_int_distribution<uint16_t> distribution(0, max_uint8);
+    auto& twister = pseudo_random::get_twister();
+
+    std::transform(out.begin(), out.end(), out.begin(),
+        [&distribution, &twister](uint8_t byte) {
+            return distribution(twister); });
+}
```
evoskuil: Style, balance braces:
```cpp
    std::transform(out.begin(), out.end(), out.begin(),
        [&distribution, &twister](uint8_t byte)
        {
            return distribution(twister);
        });
```
I prefer to write these as:
```cpp
    const auto fill = [&distribution, &twister](uint8_t byte)
    {
        return distribution(twister);
    };

    std::transform(out.begin(), out.end(), out.begin(), fill);
```

src/utility/random.cpp
```diff
@@ -101,7 +125,7 @@ asio::duration pseudo_randomize(const asio::duration& expiration,
         return expiration;
 
     // [0..2^64) % 2500 => [0..2500]
-    const auto random_offset = static_cast<int>(pseudo_random(0, limit));
+    const auto random_offset = static_cast<int>(next(0, limit));
```
evoskuil: `pseudo_random::next(` ?

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -50,10 +57,52 @@ BC_API void pseudo_random_fill(data_chunk& chunk);
...
+
+    /**
+     * Generate a pseudo random number within the domain.
+     * @return  The 64 bit number.
+     */
+    static BC_API uint64_t next();
```
evoskuil: The `BC_API` should qualify the class (except for template classes, which are embedded, not linked).

include/bitcoin/bitcoin/utility/random.hpp
evoskuil: Don't use `BC_API` on template methods (or template classes).

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -29,18 +30,24 @@ namespace libbitcoin {
 /**
  * Generate a pseudo random number within the domain.
  * @return  The 64 bit number.
+ *
+ * DEPRECATED in v3
```
evoskuil: See existing convention [here](https://github.com/libbitcoin/libbitcoin/blob/master/include/bitcoin/bitcoin/wallet/encrypted_keys.hpp#L72). No need to qualify the version since it's inherent in the branch. Note also that we don't use `BC_DEPRECATED` because it creates heavy build warnings.

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -29,18 +30,24 @@ namespace libbitcoin {
...
 BC_API uint64_t pseudo_random(uint64_t begin, uint64_t end);
 
 /**
  * Fill a buffer with randomness using the default random engine.
- * @param[in]  chunk  The buffer to fill with randomness.
+ * @param[in]  out  The datatype to fill with randomness.
+ *
+ * DEPRECATED in v3
  */
 BC_API void pseudo_random_fill(data_chunk& chunk);
```
evoskuil: chunk => out ?

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -50,10 +57,52 @@ BC_API void pseudo_random_fill(data_chunk& chunk);
...
+
+    /**
+     * Generate a pseudo random number within [begin, end].
+     * @return  The 64 bit number.
+     */
+    static BC_API uint64_t next(uint64_t begin, uint64_t end);
+
+    /**
+     * Fill a buffer with randomness using the default random engine.
+     * @param[in]  out  The datatype to fill with randomness.
```
evoskuil: @param[out]

evoskuil: > The datatype to fill with randomness.
`out` is the parameter while `Container` is the type (template parameter). I've stopped using these types of comments because they are costly to maintain and often incorrect.

src/utility/random.cpp
```diff
@@ -62,30 +77,39 @@ static std::mt19937& get_twister()
...
-        byte = static_cast<uint8_t>(distribution(get_twister()));
+asio::duration pseudo_randomize(const asio::duration& expiration,
+    uint8_t ratio)
+{
+    return pseudo_random::duration(expiration, ratio);
 }
 
 // Randomly select a time duration in the range:
 // [(expiration - expiration / ratio) .. expiration]
 // Not fully testable due to lack of random engine injection.
```
evoskuil: Note that by making the engine a member it can be overridden for test, making these methods now testable.

include/bitcoin/bitcoin/utility/random.hpp
```diff
@@ -19,6 +19,7 @@
 #ifndef LIBBITCOIN_RANDOM_HPP
 #define LIBBITCOIN_RANDOM_HPP
```
evoskuil: Style, file names should change to `pseudo_random` to match class (and the include guards above).

include/bitcoin/bitcoin/utility/random.hpp
thecodefactory: get_twister should be private?  How will it expose the generator?

thecodefactory: Ah, nevermind, I get it now :+1:

thecodefactory:
Updated.  Slowly getting better ;-)

include/bitcoin/bitcoin/utility/pseudo_random.hpp
```diff
@@ -0,0 +1,125 @@
...
+ * @param[in]  maximum  The maximum value to return.
+ * @param[in]  ratio    The determinant of the minimum duration as the inverse
+ *                      portion of the maximum duration.
+ * @return              The randomized duration.
+ */
+BC_API asio::duration pseudo_randomize(const asio::duration& maximum,
+    uint8_t ratio=2);
+
+
+class BC_API pseudo_random
```
evoskuil: Style, move to top of file (before functions).

include/bitcoin/bitcoin/utility/pseudo_random.hpp
```diff
@@ -0,0 +1,125 @@
...
+{
+  public:
+    /**
+     * Fill a container with randomness using the default random engine.
+     */
+    template<class Container>
+    static void fill(Container& out)
```
evoskuil: This will work for now, but we eventually hit problems with header inclusion of template methods.

include/bitcoin/bitcoin/utility/pseudo_random.hpp
```diff
@@ -0,0 +1,125 @@
...
+     */
+    static asio::duration duration(const asio::duration& maximum,
+        uint8_t ratio=2);
+
+  private:
+    /**
+     * Returns a thread-safe static random generator instance.
+     * @return  The generator.
+     */
+    static std::mt19937& get_twister();
```
evoskuil: Don't need doc comments for privates.

src/utility/pseudo_random.cpp
```diff
@@ -62,30 +76,27 @@ static std::mt19937& get_twister()
     return *twister;
```
evoskuil: Above comment should be:
`// This is thread safe because the instance is thread static.`
must have got lost in rebase.

evoskuil:
Looking good.

include/bitcoin/bitcoin/utility/pseudo_random.hpp
thecodefactory: Where?  Just wondering.  I have been porting the -network changes for the other PR with each revision of this and haven't hit them yet.

PR #917, opened 2018-03-06
Expose payment_record index to fix dependent break.

pmienk:


src/chain/payment_record.cpp
```diff
@@ -322,5 +322,17 @@ void payment_record::set_hash(hash_digest&& hash)
...
 
+uint32_t payment_record::index() const
+{
+    return index_;
+}
+
+void payment_record::set_index(uint32_t value)
+{
+    // This is no longer a default instance, so valid.
+    valid_ = true;
```
evoskuil: Need to apply this to `payment_record::set_hash(hash_digest&& hash)` as well for consistency.

pmienk: Noted, in following commit.

PR #925, opened 2018-03-31
Regenerate for android support.

pmienk:


narodnik:
Hello I still get the same error.

You can reproduce by installing docker (there are packages for every platform) then running these 2 commands:

`$ git clone -b nopatch https://github.com/narodnik/lbdocker.git
$ cd lbdocker
$ docker build . -t dw
`

pmienk:
I encounter the following:

```
configure: WARNING: using cross tools not prefixed with host triplet
```

I don't have any experience with docker, but it appears that you've got an image that installs cross compilation tools without properly specifying the `host` parameter.  I assume docker utilizes the host system, linux, and therefore properly requires librt and pthread.  The tools would imply to me a ```host``` containing something like ```arm-linux-androideabi``` which should pass both linux and android checks and properly exclude those dependencies if specified.

pmienk:
I see what's going on, will fix.

pmienk:
Corrected in #931, libbitcoin/libbitcoin-explorer#546, libbitcoin/libbitcoin-build#157.

PR #947, opened 2018-05-10
Ring signature

narodnik:
This adds Borromean ring signatures. The unit test is exactly the same as the experimental secp256k1-zkp branch. I will spend more time verifying and auditing this code, but it will also require some review by an outsider. I'm confident the code is better written and more understandable, more condensed than the secp256k1-zkp code. I tried to keep the interface design similar to ec sign/verify but there are some subtle differences, namely that the message digest is computed by hashing it together with the signing keys and done internally in the code.

### DIFF
--- a/include/bitcoin/bitcoin/math/elliptic_curve.hpp
+++ b/include/bitcoin/bitcoin/math/elliptic_curve.hpp
@@ -81,6 +81,14 @@ static BC_CONSTEXPR ec_uncompressed null_uncompressed_point =
 // Add and multiply EC values
 // ----------------------------------------------------------------------------
 
+/// Negate a scalar
+/// return false on failure
+BC_API bool ec_negate(ec_secret& scalar);
+
+/// Invert a point (flip on Y axis)
+/// return false on failure
+BC_API bool ec_negate(ec_compressed& point);
+
 /// Compute the sum a += G*b, where G is the curve's generator point.
 /// return false on failure (such as infinity or zero).
 BC_API bool ec_add(ec_compressed& point, const ec_secret& secret);
--- a/include/bitcoin/bitcoin/math/elliptic_curve.hpp
+++ b/include/bitcoin/bitcoin/math/elliptic_curve.hpp
@@ -34,6 +34,8 @@ static BC_CONSTEXPR uint8_t ec_even_sign = 2;
 static BC_CONSTEXPR size_t ec_secret_size = 32;
 typedef byte_array<ec_secret_size> ec_secret;
 
+typedef std::vector<ec_secret> secret_list;
+
 /// Compressed public key:
 static BC_CONSTEXPR size_t ec_compressed_size = 33;
 typedef byte_array<ec_compressed_size> ec_compressed;
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -39,7 +39,7 @@ typedef std::vector<point_list> key_rings;
 /**
  * A borromean ring signature.
  *
- * theta = {e_0, s_i_j : 0 <= i <= n, 0 <= j <= m_i}
+ * theta = {e_0, s_{i_j} : 0 <= i <= n, 0 <= j <= m_i}
  */
 struct ring_signature
 {
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -23,21 +23,51 @@
 
 namespace libbitcoin {
 
+/**
+ * key_rings is a collection of public key rings.
+ * Each ring contains several public keys.
+ * To create a valid signature, you must use at least one valid private
+ * key from each ring.
+ * That is given three rings [{A, B, C}, {D, E, F}, {X, Y}], then to
+ * create a valid signature, we must use a private key from each of those
+ * sets.
+ * For example A and E and X. We can summarize this operation as:
+ * (A or B or C) and (D or E or F) and (X or Y)
+ */
 typedef std::vector<point_list> key_rings;
+
+/**
+ * List of secret keys used for signing. There must be one valid secret
+ * key from each ring.
+ * For example given a ring of [{A, B, C}, {D, E, F}, {X, Y}] then we
+ * must have a set of keys that satisfies this constraint:
+ * (A or B or C) and (D or E or F) and (X or Y)
+ */
 typedef std::vector<ec_secret> secret_list;
 
-typedef std::vector<secret_list> s_values_type;
-
+/**
+ * A borromean ring signature.
+ *
+ * theta = {e_0, s_i_j : 0 <= i <= n, 0 <= j <= m_i}
+ */
 struct ring_signature
 {
+    typedef std::vector<secret_list> s_values_type;
+
     ec_secret e;
     s_values_type s;
 };
 
 /**
- * message: the message to sign
- * rings: an array rings; each ring is an array of pubkeys
- * secret_keys: an array of signing keys
+ * Create a borromean ring signature.
+ * There must exist a valid signing key for each ring of public keys.
+ * @param[in]  out      The new signature.
+ * @param[in]  secrets  Secret signing keys. There should be at least one key
+ *                      from each ring.
+ * @param[in]  rings    The rings each with N_i public keys.
+ * @param[in]  message  The message data to sign.
+ * @param[in]  seed     Randomizing seed data.
+ * @return false if the signing operation fails.
  */
 BC_API bool sign(
     ring_signature& out,
@@ -46,6 +76,13 @@ BC_API bool sign(
     const data_slice message,
     const data_slice seed);
 
+/**
+ * Verify a borromean ring signature.
+ * @param[in]  rings        The rings each with N_i public keys.
+ * @param[in]  message      The message data to verify.
+ * @param[in]  signature    Signature.
+ * @return false if the verify operation fails.
+ */
 BC_API bool verify(
     const key_rings& rings,
     const data_slice message,
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -24,7 +24,7 @@
 namespace libbitcoin {
 
 /**
- * key_rings is a collection of public key rings.
+ * key_rings is a collection of rings of public keys.
  * Each ring contains several public keys.
  * To create a valid signature, you must use at least one valid private
  * key from each ring.
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -77,9 +77,6 @@ BC_API bool sign(ring_signature& out, const secret_list& secrets,
     const key_rings& rings, const hash_digest& digest,
     const secret_list& salts);
 
-BC_API bool sign(ring_signature& out, const secret_list& secrets,
-    const key_rings& rings, const data_slice message, const data_slice seed);
-
 /**
  * Verify a borromean ring signature.
  *
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -65,6 +65,10 @@ struct ring_signature
  * @return false if the signing operation fails.
  */
 BC_API bool sign(ring_signature& out, const secret_list& secrets,
+    const key_rings& rings, const hash_digest& M,
+    const secret_list& k);
+
+BC_API bool sign(ring_signature& out, const secret_list& secrets,
     const key_rings& rings, const data_slice message, const data_slice seed);
 
 /**
@@ -75,6 +79,9 @@ BC_API bool sign(ring_signature& out, const secret_list& secrets,
  * @param[in]  signature    Signature.
  * @return false if the verify operation fails.
  */
+BC_API bool verify(const key_rings& rings, const hash_digest& M,
+    const ring_signature& signature);
+
 BC_API bool verify(const key_rings& rings, const data_slice message,
     const ring_signature& signature);
 
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -59,7 +59,7 @@ struct ring_signature
  * @param[in]  out      The new signature.
  * @param[in]  secrets  Secret signing keys. There should be at least one key
  *                      from each ring.
- * @param[in]  rings    The rings each with N_i public keys.
+ * @param[in]  rings    The rings each with n_i public keys.
  * @param[in]  message  The message data to sign.
  * @param[in]  seed     Randomizing seed data.
  * @return false if the signing operation fails.
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -43,11 +43,20 @@ typedef std::vector<point_list> key_rings;
  */
 struct ring_signature
 {
-    typedef std::vector<secret_list> s_values_type;
+    typedef std::vector<secret_list> proof_list;
 
-    ec_secret e;
-    s_values_type s;
+    // The initial 'e' value
+    ec_secret challenge;
+    // The 's' values
+    proof_list proofs;
 };
+
+/**
+ * Prepare hash digest for use in ring signature signing and verification.
+ * Returns sha256(message || flatten(rings))
+ * @return     digest   Ring signature digest
+ */
+BC_API hash_digest prepare_digest(data_slice message, const key_rings& rings);
 
 /**
  * Create a borromean ring signature.
@@ -60,13 +69,13 @@ struct ring_signature
  * @param[in]  secrets  Secret signing keys. There should be at least one key
  *                      from each ring.
  * @param[in]  rings    The rings each with n_i public keys.
- * @param[in]  message  The message data to sign.
- * @param[in]  seed     Randomizing seed data.
+ * @param[in]  digest   The message digest to sign.
+ * @param[in]  salts    Randomizing seed values.
  * @return false if the signing operation fails.
  */
 BC_API bool sign(ring_signature& out, const secret_list& secrets,
-    const key_rings& rings, const hash_digest& M,
-    const secret_list& k);
+    const key_rings& rings, const hash_digest& digest,
+    const secret_list& salts);
 
 BC_API bool sign(ring_signature& out, const secret_list& secrets,
     const key_rings& rings, const data_slice message, const data_slice seed);
@@ -75,14 +84,11 @@ BC_API bool sign(ring_signature& out, const secret_list& secrets,
  * Verify a borromean ring signature.
  *
  * @param[in]  rings        The rings each with N_i public keys.
- * @param[in]  message      The message data to verify.
+ * @param[in]  digest       The message digest to verify.
  * @param[in]  signature    Signature.
  * @return false if the verify operation fails.
  */
-BC_API bool verify(const key_rings& rings, const hash_digest& M,
-    const ring_signature& signature);
-
-BC_API bool verify(const key_rings& rings, const data_slice message,
+BC_API bool verify(const key_rings& rings, const hash_digest& digest,
     const ring_signature& signature);
 
 } // namespace libbitcoin
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -1,0 +1,57 @@
+/**
+ * Copyright (c) 2011-2018 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_RING_SIGNATURE_HPP
+#define LIBBITCOIN_RING_SIGNATURE_HPP
+
+#include <bitcoin/bitcoin/math/elliptic_curve.hpp>
+
+namespace libbitcoin {
+
+typedef std::vector<point_list> key_rings;
+typedef std::vector<ec_secret> secret_list;
+
+typedef std::vector<secret_list> s_values_type;
+
+struct ring_signature
+{
+    ec_secret e;
+    s_values_type s;
+};
+
+/**
+ * message: the message to sign
+ * rings: an array rings; each ring is an array of pubkeys
+ * secret_keys: an array of signing keys
+ */
+BC_API bool sign(
+    ring_signature& out,
+    const secret_list& secrets,
+    const key_rings& rings,
+    const data_slice message,
+    const data_slice seed);
+
+BC_API bool verify(
+    const key_rings& rings,
+    const data_slice message,
+    const ring_signature& signature);
+
+} // namespace libbitcoin
+
+#endif
+
--- a/include/bitcoin/bitcoin/math/ring_signature.hpp
+++ b/include/bitcoin/bitcoin/math/ring_signature.hpp
@@ -37,15 +37,6 @@ namespace libbitcoin {
 typedef std::vector<point_list> key_rings;
 
 /**
- * List of secret keys used for signing. There must be one valid secret
- * key from each ring.
- * For example given a ring of [{A, B, C}, {D, E, F}, {X, Y}] then we
- * must have a set of keys that satisfies this constraint:
- * (A or B or C) and (D or E or F) and (X or Y)
- */
-typedef std::vector<ec_secret> secret_list;
-
-/**
  * A borromean ring signature.
  *
  * theta = {e_0, s_i_j : 0 <= i <= n, 0 <= j <= m_i}
@@ -61,6 +52,10 @@ struct ring_signature
 /**
  * Create a borromean ring signature.
  * There must exist a valid signing key for each ring of public keys.
+ * For example given a ring of [{A, B, C}, {D, E, F}, {X, Y}] then we
+ * must have a set of keys that satisfies this constraint:
+ * (A or B or C) and (D or E or F) and (X or Y)
+ *
  * @param[in]  out      The new signature.
  * @param[in]  secrets  Secret signing keys. There should be at least one key
  *                      from each ring.
@@ -69,23 +64,18 @@ struct ring_signature
  * @param[in]  seed     Randomizing seed data.
  * @return false if the signing operation fails.
  */
-BC_API bool sign(
-    ring_signature& out,
-    const secret_list& secrets,
-    const key_rings& rings,
-    const data_slice message,
-    const data_slice seed);
+BC_API bool sign(ring_signature& out, const secret_list& secrets,
+    const key_rings& rings, const data_slice message, const data_slice seed);
 
 /**
  * Verify a borromean ring signature.
+ *
  * @param[in]  rings        The rings each with N_i public keys.
  * @param[in]  message      The message data to verify.
  * @param[in]  signature    Signature.
  * @return false if the verify operation fails.
  */
-BC_API bool verify(
-    const key_rings& rings,
-    const data_slice message,
+BC_API bool verify(const key_rings& rings, const data_slice message,
     const ring_signature& signature);
 
 } // namespace libbitcoin
--- a/include/bitcoin/bitcoin/utility/serializer.hpp
+++ b/include/bitcoin/bitcoin/utility/serializer.hpp
@@ -81,9 +81,6 @@ public:
     void write_byte(uint8_t value);
 
     /// Write all bytes.
-    void write_bytes(const data_chunk& data);
-
-    /// Write all bytes.
     void write_bytes(const data_slice data);
 
     /// Write required size buffer.
--- a/include/bitcoin/bitcoin/utility/serializer.hpp
+++ b/include/bitcoin/bitcoin/utility/serializer.hpp
@@ -83,6 +83,9 @@ public:
     /// Write all bytes.
     void write_bytes(const data_chunk& data);
 
+    /// Write all bytes.
+    void write_bytes(const data_slice data);
+
     /// Write required size buffer.
     void write_bytes(const uint8_t* data, size_t size);
 
--- a/src/math/ring_signature.cpp
+++ b/src/math/ring_signature.cpp
@@ -28,16 +28,18 @@ namespace libbitcoin {
 
 typedef std::map<ec_compressed, ec_secret> secret_keys_map;
 
+typedef std::vector<size_t> index_list;
+
 // Take a list of secret keys and generate a mapping from public key -> secret
-const secret_keys_map generate_keys_map(const secret_list& secrets)
+secret_keys_map generate_keys_map(const secret_list& secrets)
 {
     secret_keys_map keys;
     for (const auto& secret: secrets)
     {
         // Compute public key
         ec_compressed public_key;
-        bool rc = secret_to_public(public_key, secret);
-        BITCOIN_ASSERT(rc);
+        if (!secret_to_public(public_key, secret))
+            return secret_keys_map();
 
         // Insert into map
         keys.insert({ public_key, secret });
@@ -49,8 +51,7 @@ const secret_keys_map generate_keys_map(const secret_list& secrets)
 // Make a list of public keys for which we have the corresponding secret key
 // in a single ring of public keys.
 const point_list known_keys_in_ring(
-    const secret_keys_map& secret_keys,
-    const point_list& ring)
+    const secret_keys_map& secret_keys, const point_list& ring)
 {
     point_list known_ring;
     for (const auto& key: ring)
@@ -64,8 +65,7 @@ const point_list known_keys_in_ring(
 
 // For all rings, make a list of known public keys corresponding to each ring.
 const key_rings partition_keys_into_rings(
-    const secret_keys_map& secret_keys,
-    const key_rings& rings)
+    const secret_keys_map& secret_keys, const key_rings& rings)
 {
     key_rings known_keys;
     // For each ring in our rings...
@@ -79,44 +79,32 @@ const key_rings partition_keys_into_rings(
     return known_keys;
 }
 
-// Check that all rings are not empty
-bool all_rings_nonzero_size(const key_rings& known_keys)
-{
-    for (const auto& ring: known_keys)
-        if (ring.empty())
-            return false;
-    return true;
-}
-
-typedef std::vector<size_t> index_list;
-
 // Make a list of indexes of where our known key occurs in each ring of
 // public keys. That is given a ring of {A, B, C} where we know the
 // private key of B, it will return 1 (the index in the ring).
 // This function computes this for all rings.
 index_list search_key_indexes(
-    const key_rings& rings,
-    const key_rings& known_keys_by_ring)
+    const key_rings& rings, const key_rings& known_keys_by_ring)
 {
     index_list known_key_indexes;
+    BITCOIN_ASSERT(known_keys_by_ring.size() == rings.size());
     // Iterate all our rings.
     for (size_t i = 0; i < rings.size(); ++i)
     {
-        BITCOIN_ASSERT(i < known_keys_by_ring.size());
         const auto& ring = rings[i];
         const auto& known = known_keys_by_ring[i];
         // Find known key in this current ring.
         const auto it = std::find(ring.begin(), ring.end(), known.back());
         BITCOIN_ASSERT(it != ring.end());
         // Calculate and store the index
-        size_t index = it - ring.begin();
+        const auto index = std::distance(ring.begin(), it);
         known_key_indexes.push_back(index);
     }
     return known_key_indexes;
 }
 
 // Returns message || flatten(rings)
-data_chunk concatenate(const data_slice message, const key_rings& rings)
+data_chunk concatenate(data_slice message, const key_rings& rings)
 {
     data_chunk result(message.begin(), message.end());
     for (const auto& ring: rings)
@@ -130,17 +118,17 @@ data_chunk concatenate(const data_slice message, const key_rings& rings)
 // Each key in each ring, has a corresponding s value.
 // That is, each ring[i][j] should have a corresponding s[i][j]
 void random_k_and_s_values(
-    secret_list& k, ring_signature::s_values_type& s,
-    const key_rings& rings, const data_slice seed)
+    secret_list& out_k, ring_signature::s_values_type& out_s,
+    const key_rings& rings, data_slice seed)
 {
     // This value increments inside this function with every call
     // to random_scalar() lambda.
-    uint32_t increment_randomizer = 0;
+    uint32_t nonce = 0;
     // Use a hd_private wallet as a random generator for ec_secret values.
-    auto random_scalar = [&seed, &increment_randomizer]()
+    const auto random_scalar = [](data_slice seed, uint32_t& nonce)
     {
         static const wallet::hd_private generate_secret(to_chunk(seed));
-        return generate_secret.derive_private(increment_randomizer++).secret();
+        return generate_secret.derive_private(nonce++).secret();
     };
 
     // Each ring[i][j] should have a corresponding s[i][j]
@@ -148,10 +136,10 @@ void random_k_and_s_values(
     {
         secret_list s_ring;
         for (const auto& key: ring)
-            s_ring.push_back(random_scalar());
-        s.push_back(s_ring);
+            s_ring.push_back(random_scalar(seed, nonce));
+        out_s.push_back(s_ring);
 
-        k.push_back(random_scalar());
+        out_k.push_back(random_scalar(seed, nonce));
     }
 }
 
@@ -203,34 +191,41 @@ ec_secret calculate_s(const ec_secret& k, const ec_secret& e,
     rc = ec_negate(result);
     BITCOIN_ASSERT(rc);
     // result = k - e x
-    ec_add(result, k);
+    rc = ec_add(result, k);
+    BITCOIN_ASSERT(rc);
     return result;
 }
 
-bool sign(
-    ring_signature& out,
-    const secret_list& secrets,
-    const key_rings& rings,
-    const data_slice message,
-    const data_slice seed)
+bool sign(ring_signature& out, const secret_list& secrets,
+    const key_rings& rings, const data_slice message, const data_slice seed)
 {
     // Create public key -> secret key map
     auto secret_keys = generate_keys_map(secrets);
+    if (secret_keys.empty())
+        return false;
     // Organize known public keys into corresponding rings
     const auto known_keys_by_ring =
         partition_keys_into_rings(secret_keys, rings);
+
     // Check we know a secret key in each ring
-    if (!all_rings_nonzero_size(known_keys_by_ring))
+    auto non_empty_ring = [](const point_list& ring)
+    {
+        return !ring.empty();
+    };
+    const bool all_rings_nonzero_size = std::all_of(
+        known_keys_by_ring.begin(), known_keys_by_ring.end(), non_empty_ring);
+
+    if (!all_rings_nonzero_size)
         return false;
 
     // Compute indexes for known keys inside the rings.
-    const index_list known_key_indexes =
+    const auto known_key_indexes =
         search_key_indexes(rings, known_keys_by_ring);
     BITCOIN_ASSERT(known_key_indexes.size() == rings.size());
 
     // Compute message digest M
-    const data_chunk message_data = concatenate(message, rings);
-    const hash_digest M = sha256_hash(message_data);
+    const auto message_data = concatenate(message, rings);
+    const auto M = sha256_hash(message_data);
 
     // Generate random k and s values
     secret_list k;
--- a/src/math/ring_signature.cpp
+++ b/src/math/ring_signature.cpp
@@ -295,6 +295,29 @@ bool sign(ring_signature& out, const secret_list& secrets,
     return true;
 }
 
+bool compute_last_R(ec_compressed& R_i_j, const point_list& ring,
+    ec_secret e_i_j, size_t i, const hash_digest& digest,
+    const ring_signature& signature)
+{
+    BITCOIN_ASSERT(signature.proofs[i].size() == ring.size());
+    for (size_t j = 0; j < ring.size(); ++j)
+    {
+        // s_i_j
+        const auto& s = signature.proofs[i][j];
+
+        if (is_zero(s) || is_zero(e_i_j))
+            return false;
+
+        // Calculate R and e values until the end.
+        if (!calculate_R(R_i_j, s, e_i_j, ring[j]))
+            return false;
+
+        // Next e value
+        e_i_j = borromean_hash(digest, R_i_j, i, j + 1);
+    }
+    return true;
+}
+
 bool verify(const key_rings& rings, const hash_digest& digest,
     const ring_signature& signature)
 {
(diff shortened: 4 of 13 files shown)


src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const auto& ring = rings[i];
+        const size_t known_key_index = known_key_indexes[i];
+
+        // Calculate starting e value of this current ring.
+        auto e_i_j = borromean_hash(M, out.e, i, 0);
+
+        // Loop until index of known key.
+        for (size_t j = 0; j < known_key_index; ++j)
+        {
+            BITCOIN_ASSERT(j < out.s[i].size());
```
evoskuil: Test outside of loop. Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+            const auto& s = out.s[i][j];
+
+            // Calculate e and R until we reach our index.
+            BITCOIN_ASSERT(j < ring.size());
```
evoskuil: Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+            const auto R_i_j = calculate_R(s, e_i_j, ring[j]);
+            e_i_j = borromean_hash(M, R_i_j, i, j + 1);
+        }
+
+        // Find secret key used for calculation in the next step.
+        BITCOIN_ASSERT(known_key_index < ring.size());
```
evoskuil: Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const auto& known_public_key = ring[known_key_index];
+        BITCOIN_ASSERT(secret_keys.find(known_public_key) != secret_keys.end());
```
evoskuil: Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // it by calculating e0, and looping all the way until the end R value
+    // which we use to recalculate e0.
+    // If the values match then we have a valid ring signature.
+
+    data_chunk e0_data;
+
+    // Loop through rings
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < signature.s.size());
```
evoskuil: Test outside of loop. Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+        const auto& ring = rings[i];
+
+        // Calculate first e value for this ring.
+        auto e_i_j = borromean_hash(M, signature.e, i, 0);
+
+        ec_compressed R_i_j;
+        for (size_t j = 0; j < ring.size(); ++j)
+        {
+            BITCOIN_ASSERT(j < signature.s[i].size());
```
evoskuil: Test outside of loop. Is this assertion *possible*?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+            const auto& s = signature.s[i][j];
+
+            // Calculate R and e values until the end.
+            BITCOIN_ASSERT(j < ring.size());
```
evoskuil: This assertion is unnecessary (it's the loop condition).

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    const data_chunk message_data = concatenate(message, rings);
+    const hash_digest M = sha256_hash(message_data);
+
+    // As compared with signing, we only have to perform a single step.
+    // The ring has already been computed, so now we just need to verify
+    // it by calculating e0, and looping all the way until the end R value
+    // which we use to recalculate e0.
+    // If the values match then we have a valid ring signature.
+
+    data_chunk e0_data;
```
evoskuil: Reservable?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+            // Calculate R and e values until the end.
+            BITCOIN_ASSERT(j < ring.size());
+            R_i_j = calculate_R(s, e_i_j, ring[j]);
+            e_i_j = borromean_hash(M, R_i_j, i, j + 1);
+        }
+        extend_data(e0_data, R_i_j);
+    }
+    extend_data(e0_data, M);
+    // Hash data to produce e0 value
+    auto e0_hash = sha256_hash(e0_data);
```
evoskuil: Style: use `const`.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+bool sign(
+    ring_signature& out,
+    const secret_list& secrets,
+    const key_rings& rings,
+    const data_slice message,
+    const data_slice seed)
+{
+    // Create public key -> secret key map
+    auto secret_keys = generate_keys_map(secrets);
```
evoskuil: Style: use `const`.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    return result;
+}
+
+// Generate random k and s values.
+// There is a single k for each ring.
+// Each key in each ring, has a corresponding s value.
+// That is, each ring[i][j] should have a corresponding s[i][j]
+void random_k_and_s_values(
+    secret_list& k, ring_signature::s_values_type& s,
+    const key_rings& rings, const data_slice seed)
```
evoskuil: Remove `const` on `data_slice`.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+            extend_data(result, key);
+    return result;
+}
+
+// Generate random k and s values.
+// There is a single k for each ring.
+// Each key in each ring, has a corresponding s value.
+// That is, each ring[i][j] should have a corresponding s[i][j]
+void random_k_and_s_values(
+    secret_list& k, ring_signature::s_values_type& s,
```
evoskuil: Prefer `out_k` and `out_s` here.

evoskuil: Style: don't wrap parameters until hit 80 char limit.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    const key_rings& rings, const data_slice seed)
+{
+    // This value increments inside this function with every call
+    // to random_scalar() lambda.
+    uint32_t increment_randomizer = 0;
+    // Use a hd_private wallet as a random generator for ec_secret values.
+    auto random_scalar = [&seed, &increment_randomizer]()
```
evoskuil: Style: use `const`.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        static const wallet::hd_private generate_secret(to_chunk(seed));
+        return generate_secret.derive_private(increment_randomizer++).secret();
+    };
+
+    // Each ring[i][j] should have a corresponding s[i][j]
+    for (const auto& ring: rings)
+    {
+        secret_list s_ring;
+        for (const auto& key: ring)
+            s_ring.push_back(random_scalar());
```
evoskuil: The code is more complicated by use of a closure than a utility function or a closure that doesn't capture, and the `key` reference isn't used. Suggest:
```cpp
    for (size_t nonce; nonce < ring.size(); ++nonce)
        s_ring.push_back(random_scalar(nonce, seed));
```

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+}
+
+bool verify(
+    const key_rings& rings,
+    const data_slice message,
+    const ring_signature& signature)
+{
+    // Compute message digest M
+    const data_chunk message_data = concatenate(message, rings);
+    const hash_digest M = sha256_hash(message_data);
```
evoskuil: Style: use `auto` for both lines.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+#include <random>
+#include <boost/test/unit_test.hpp>
+#include <bitcoin/bitcoin.hpp>
+
+using namespace bc;
+
+data_chunk new_seed()
+{
+    data_chunk seed(10);
+    pseudo_random_fill(seed);
```
evoskuil: Don't use dynamically-selected test vectors.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    return seed;
+}
+
+ec_secret new_key()
+{
+    const data_chunk seed = new_seed();
```
evoskuil: Style: use `auto`.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+template <typename SecretsIterator>
+point_list extract_public_keys(SecretsIterator start, SecretsIterator end)
+{
+    point_list ring;
+    for (auto it = start; it != end; ++it)
+    {
+        ec_compressed public_key;
+        bool rc = secret_to_public(public_key, *it);
+        BITCOIN_ASSERT(rc);
```
evoskuil: Remove result suppression (factored out of NDEBUG builds)) and rely on test helpers.
```cpp
BOOST_REQUIRE(secret_to_public(public_key, *it));
```

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+        ring.push_back(public_key);
+    }
+    return ring;
+}
+
+bool is_secret_in_ring(const ec_secret& secret, const point_list& ring)
+{
+    ec_compressed public_key;
+    bool rc = secret_to_public(public_key, secret);
+    BITCOIN_ASSERT(rc);
```
evoskuil: Use `BOOST_REQUIRE`.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    for (const auto& ring_key: ring)
+    {
+        if (ring_key == public_key)
+            return true;
```
evoskuil: use `std::any_of`.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    // Generate new random secret keys
+    secret_list secrets;
+    for (size_t i = 0; i < 8; ++i)
+        secrets.push_back(new_key());
+
+    // Create two rings from the first 4 and last 4 keys
+    auto ring0 = extract_public_keys(
+        secrets.begin(), secrets.begin() + 4);
+    auto ring1 = extract_public_keys(
+        secrets.begin() + 4, secrets.end());
```
evoskuil: Avoid wrapping lines in test cases (relaxed requirement vs. non-test code).
The validity of the test can only be verified visually, so it must be a simple series of declarations.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    {
+        if (ring_key == public_key)
+            return true;
+    }
+    return false;
+}
+
+BOOST_AUTO_TEST_SUITE(ring_signature_tests)
+
+BOOST_AUTO_TEST_CASE(ring_signature__basic_test)
```
evoskuil: One giant test case is not ideal. Each individual function/method should be tested independently for each of its conditions. That allows us to see exactly what is broken when a test fails. This is okay as an additional test if it's meant to duplicate a reference implementation.

Test cases should be fully declarative, without internal loops or conditions. There are a few exceptions in libbitcoin, some for historical reasons. 

See also the style/objective of test naming: https://github.com/libbitcoin/libbitcoin/wiki/Test-Guide

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+{
+    // Generate new random secret keys
+    secret_list secrets;
+    for (size_t i = 0; i < 8; ++i)
```
evoskuil: Remove loop in favor of fixed test vector.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    std::string message = "hello world!";
+    data_chunk message_data = to_chunk(message);
+
+    ring_signature signature;
+    const auto seed = new_seed();
+    BOOST_REQUIRE(sign(signature, secret_keys, rings, message_data, seed));
+
+    BOOST_REQUIRE(verify(rings, message_data, signature));
+
+    for (size_t i = 0; i < rings.size(); ++i)
```
evoskuil: Remove loops to test utility code.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+        secrets.begin() + 4, secrets.end());
+    BOOST_REQUIRE_EQUAL(ring0.size(), 4);
+    BOOST_REQUIRE_EQUAL(ring1.size(), 4);
+    const key_rings rings{ring0, ring1};
+
+    // Randomly choose two random secret keys which exist in both rings
+    // So one key from first 4, and one from last 4
+    secret_list secret_keys;
+    secret_keys.push_back(secrets[pseudo_random(0, 3)]);
+    secret_keys.push_back(secrets[pseudo_random(4, 7)]);
```
evoskuil: Don't use random vectors.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BOOST_REQUIRE(i < signature.s.size());
+        const auto& ring = rings[i];
+        for (size_t j = 0; j < ring.size(); ++j)
+        {
+            BOOST_REQUIRE(j < signature.s[i].size());
+            ring_signature signature_copy = signature;
+            if (pseudo_random(0, 1))
```
evoskuil: Don't use random vectors.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+                ec_negate(signature_copy.s[i][j]);
+            else
+            {
+                ec_secret one;
+                std::fill(one.begin(), one.end(), 0);
+                one.back() = 1;
```
evoskuil: Declare `one` as a static constant outside the test case/loop. Can use `hash_literal`.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+    BOOST_REQUIRE(verify(rings, message_data, signature));
+
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BOOST_REQUIRE(i < signature.s.size());
+        const auto& ring = rings[i];
+        for (size_t j = 0; j < ring.size(); ++j)
+        {
+            BOOST_REQUIRE(j < signature.s[i].size());
+            ring_signature signature_copy = signature;
```
evoskuil: Style: use `auto`.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+            if (pseudo_random(0, 1))
+                ec_negate(signature_copy.s[i][j]);
```
evoskuil: Don't suppress error cases: `BOOST_REQUIRE(ec_negate(...))`

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+            else
+            {
+                ec_secret one;
+                std::fill(one.begin(), one.end(), 0);
+                one.back() = 1;
+                ec_add(signature_copy.s[i][j], one);
```
evoskuil: Don't suppress error cases: `BOOST_REQUIRE(ec_add(...))`

evoskuil:
```
  CXX      src/math/src_libbitcoin_la-ring_signature.lo
src/math/elliptic_curve.cpp:73:9: error: use of undeclared identifier 'secp256k1_ec_pubkey_negate'
        secp256k1_ec_pubkey_negate(context, &pubkey) == 1 &&
```
The build failure likely results from missing this include:
```
#include <secp256k1.h>
```

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    secret_list& k, ring_signature::s_values_type& s,
+    const key_rings& rings, const data_slice seed)
+{
+    // This value increments inside this function with every call
+    // to random_scalar() lambda.
+    uint32_t increment_randomizer = 0;
+    // Use a hd_private wallet as a random generator for ec_secret values.
+    auto random_scalar = [&seed, &increment_randomizer]()
+    {
+        static const wallet::hd_private generate_secret(to_chunk(seed));
```
evoskuil: I'd prefer to not have a dependency on `bc::wallet` here. It might be worth at some point factoring out a seed->ec_secret function and placing into elliptic_curve.cpp.

include/bitcoin/bitcoin/utility/serializer.hpp
```diff
@@ -83,6 +83,9 @@ class serializer
     /// Write all bytes.
     void write_bytes(const data_chunk& data);
 
+    /// Write all bytes.
+    void write_bytes(const data_slice data);
```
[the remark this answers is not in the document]
narodnik: As you prefer. I thought maybe in some low level db code, that the pointer misdirection might result in cache misses but maybe that premature optimization.

include/bitcoin/bitcoin/math/elliptic_curve.hpp
```diff
@@ -81,6 +81,14 @@ static BC_CONSTEXPR ec_uncompressed null_uncompressed_point =
...
 // ----------------------------------------------------------------------------
 
+/// Negate a scalar
+/// return false on failure
+BC_API bool ec_negate(ec_secret& scalar);
+
+/// Invert a point (flip on Y axis)
+/// return false on failure
+BC_API bool ec_negate(ec_compressed& point);
+
```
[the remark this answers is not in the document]
narodnik: I was following the order in the secp256k1.h header. Which do you prefer?

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+{
+    secret_keys_map keys;
+    for (const auto& secret: secrets)
+    {
+        // Compute public key
+        ec_compressed public_key;
+        bool rc = secret_to_public(public_key, secret);
+        BITCOIN_ASSERT(rc);
+        // Insert into map
+        keys.insert({public_key, secret});
```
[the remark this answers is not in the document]
narodnik: Understood

test/math/ring_signature.cpp
narodnik: Understood

test/math/ring_signature.cpp
narodnik: Understood

include/bitcoin/bitcoin/math/elliptic_curve.hpp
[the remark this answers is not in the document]
evoskuil: In this case ours.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+
+    ring_signature signature;
+    const auto seed = new_seed();
+    BOOST_REQUIRE(sign(signature, secret_keys, rings, message_data, seed));
+
+    BOOST_REQUIRE(verify(rings, message_data, signature));
+
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BOOST_REQUIRE(i < signature.s.size());
```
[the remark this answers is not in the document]
narodnik: You misread this here?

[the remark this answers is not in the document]
narodnik: OK I understand now

include/bitcoin/bitcoin/math/elliptic_curve.hpp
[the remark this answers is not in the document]
evoskuil: But obviously not a big deal either way, just a preference.

test/math/ring_signature.cpp
```diff
@@ -0,0 +1,126 @@
...
+        const auto& ring = rings[i];
+        for (size_t j = 0; j < ring.size(); ++j)
+        {
+            BOOST_REQUIRE(j < signature.s[i].size());
```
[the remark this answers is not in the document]
narodnik: Understood

src/math/ring_signature.cpp
narodnik: Yes, it's a bit hackish and not nice

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
evoskuil: There would be no issue with data consistency, I think I just avoided `data_slice` here for store perf reasons. But I avoid having redundant `data_chunk` and `data_slice` overloads.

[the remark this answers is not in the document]
evoskuil: Note that this also affects the `writer` interface.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+{
+    key_rings known_keys;
+    // For each ring in our rings...
+    for (const auto& ring: rings)
+    {
+        // ... find known keys corresponding to our private keys
+        const auto known_ring = known_keys_in_ring(secret_keys, ring);
+        // And add this list to our rings of known public keys.
+        known_keys.push_back(known_ring);
+    }
```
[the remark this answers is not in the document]
narodnik: Order is necessary here and must be maintained. The index of a key must match the same as the rings variable.

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
evoskuil: And therefore also the `ostream_writer` (and stream tests).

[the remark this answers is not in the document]
evoskuil: I don't think there is any perf difference at all however.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,364 @@
...
+    const hash_digest M = sha256_hash(message_data);
+
+    // Generate random k and s values
+    secret_list k;
+    random_k_and_s_values(k, out.s, rings, seed);
+
+    // ---------------------------------------------------------------------
+    // Step 1: calculate e0
+    // ---------------------------------------------------------------------
+    data_chunk e0_data;
```
evoskuil: Can this be reserved?

src/math/ring_signature.cpp
evoskuil: Note also that key generation can fail so there error must be handled (more info in bip32 and libsecp256k1).

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,84 @@
...
+typedef std::vector<point_list> key_rings;
+
+/**
+ * A borromean ring signature.
+ *
+ * theta = {e_0, s_i_j : 0 <= i <= n, 0 <= j <= m_i}
+ */
+struct ring_signature
+{
+    typedef std::vector<secret_list> s_values_type;
```
evoskuil: Use existing typedef, avoid "_type" in type names and the suffix "_values" isn't much helpful.
```cpp
struct ring_signature
{
    ec_secret e;
    secret_list s;
};
```

The cryptic single letter variable names are acceptable within the internals of the implementation (only if they are helpful in correlating to some standard), but should be avoided in public interfaces. Assume people don't know what "e" or "s" mean and replace with informative names.

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+    ec_secret e;
+    s_values_type s;
+};
```
evoskuil: Use existing typedef, avoid "_type" in type names and the suffix "_values" isn't much helpful.
```cpp
struct ring_signature
{
    ec_secret e;
    secret_list s;
};
```

The cryptic single letter variable names are acceptable within the internals of the implementation (only if they are helpful in correlating to some standard), but should be avoided in public interfaces. Assume people don't know what "e" or "s" mean and replace with informative names.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    // Loop through rings
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < signature.s.size());
+
+        const auto& ring = rings[i];
+
+        // Calculate first e value for this ring.
+        auto e_i_j = borromean_hash(M, signature.e, i, 0);
```
pmienk: From the paper (unknown if correct), e_{i, 0} = e_{0} which is a parameter of \theta equivalent to signature.e.  If that's correct, is there a reason to rehash instead of copy signature.e?

include/bitcoin/bitcoin/math/ring_signature.hpp
pmienk: I believe s is a list of lists, not the secret list.  it would be best to identify s_values_type as instead a secret_rings or equivalent to the key_rings.  Maybe public/private or other consistent naming thereof.

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
narodnik: ok changing to data_slice

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+// Take a list of secret keys and generate a mapping from public key -> secret
+const secret_keys_map generate_keys_map(const secret_list& secrets)
+{
+    secret_keys_map keys;
+    for (const auto& secret: secrets)
+    {
+        // Compute public key
+        ec_compressed public_key;
+        bool rc = secret_to_public(public_key, secret);
+        BITCOIN_ASSERT(rc);
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        // Insert into map
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+#include <boost/range/combine.hpp>
+#include <bitcoin/bitcoin/utility/serializer.hpp>
+#include <bitcoin/bitcoin/wallet/hd_private.hpp>
+
+namespace libbitcoin {
+
+typedef std::map<ec_compressed, ec_secret> secret_keys_map;
+
+// Take a list of secret keys and generate a mapping from public key -> secret
+const secret_keys_map generate_keys_map(const secret_list& secrets)
```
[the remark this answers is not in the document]
narodnik: ok I thought the compiler makes optimizations based off const, but I've removed the qualifier

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    }
+    return known_keys;
+}
+
+// Check that all rings are not empty
+bool all_rings_nonzero_size(const key_rings& known_keys)
+{
+    for (const auto& ring: known_keys)
+        if (ring.empty())
+            return false;
```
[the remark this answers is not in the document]
narodnik: thanks for the heads up

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        keys.insert({public_key, secret});
+    }
+    return keys;
+}
+
+// Make a list of public keys for which we have the corresponding secret key
+// in a single ring of public keys.
+const point_list known_keys_in_ring(
+    const secret_keys_map& secret_keys,
+    const point_list& ring)
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+// Check that all rings are not empty
+bool all_rings_nonzero_size(const key_rings& known_keys)
+{
+    for (const auto& ring: known_keys)
+        if (ring.empty())
+            return false;
+    return true;
+}
+
+typedef std::vector<size_t> index_list;
```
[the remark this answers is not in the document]
narodnik: done, good point

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    auto secret_keys = generate_keys_map(secrets);
+    // Organize known public keys into corresponding rings
+    const auto known_keys_by_ring =
+        partition_keys_into_rings(secret_keys, rings);
+    // Check we know a secret key in each ring
+    if (!all_rings_nonzero_size(known_keys_by_ring))
+        return false;
+
+    // Compute indexes for known keys inside the rings.
+    const index_list known_key_indexes =
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < known_keys_by_ring.size());
+        const auto& ring = rings[i];
+        const auto& known = known_keys_by_ring[i];
+        // Find known key in this current ring.
+        const auto it = std::find(ring.begin(), ring.end(), known.back());
+        BITCOIN_ASSERT(it != ring.end());
+        // Calculate and store the index
+        size_t index = it - ring.begin();
```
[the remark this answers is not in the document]
narodnik: ok thanks

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+// This function computes this for all rings.
+index_list search_key_indexes(
+    const key_rings& rings,
+    const key_rings& known_keys_by_ring)
+{
+    index_list known_key_indexes;
+    // Iterate all our rings.
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < known_keys_by_ring.size());
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+// public keys. That is given a ring of {A, B, C} where we know the
+// private key of B, it will return 1 (the index in the ring).
+// This function computes this for all rings.
+index_list search_key_indexes(
+    const key_rings& rings,
+    const key_rings& known_keys_by_ring)
+{
+    index_list known_key_indexes;
+    // Iterate all our rings.
+    for (size_t i = 0; i < rings.size(); ++i)
```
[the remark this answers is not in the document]
narodnik: This is the actual variable name specified in the paper on borromean signatures, and it is not the ring. Leaving this as is.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        BITCOIN_ASSERT(it != ring.end());
+        // Calculate and store the index
+        size_t index = it - ring.begin();
+        known_key_indexes.push_back(index);
+    }
+    return known_key_indexes;
+}
+
+// Returns message || flatten(rings)
+data_chunk concatenate(const data_slice message, const key_rings& rings)
```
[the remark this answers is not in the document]
narodnik: ok

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+{
+    data_chunk result(message.begin(), message.end());
+    for (const auto& ring: rings)
+        for (const auto& key: ring)
+            extend_data(result, key);
```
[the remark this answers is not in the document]
narodnik: That means I have to iterate once to calculate the size, then iterate again to build the data which will obfuscate the code.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // result = e * x
+    bool rc = ec_multiply(result, secret);
+    BITCOIN_ASSERT(rc);
+    // result = - e x
+    rc = ec_negate(result);
+    BITCOIN_ASSERT(rc);
+    // result = k - e x
+    ec_add(result, k);
+    return result;
+}
```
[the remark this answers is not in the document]
narodnik: thanks I missed this

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        return false;
+
+    // Compute indexes for known keys inside the rings.
+    const index_list known_key_indexes =
+        search_key_indexes(rings, known_keys_by_ring);
+    BITCOIN_ASSERT(known_key_indexes.size() == rings.size());
+
+    // Compute message digest M
+    const data_chunk message_data = concatenate(message, rings);
+    const hash_digest M = sha256_hash(message_data);
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < known_key_indexes.size());
+        BITCOIN_ASSERT(i < k.size());
+        BITCOIN_ASSERT(i < out.s.size());
+
+        // Current ring and index of known key
+        const auto& ring = rings[i];
+        const size_t known_key_index = known_key_indexes[i];
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    // ---------------------------------------------------------------------
+    // Step 1: calculate e0
+    // ---------------------------------------------------------------------
+    data_chunk e0_data;
+
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < known_key_indexes.size());
+        BITCOIN_ASSERT(i < k.size());
+        BITCOIN_ASSERT(i < out.s.size());
```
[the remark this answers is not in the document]
narodnik: moved them out of the loop

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+        const size_t known_key_index = known_key_indexes[i];
+
+        // Calculate starting R value...
+        ec_compressed R_i_j;
+        bool rc = secret_to_public(R_i_j, k[i]);
+
+        // ... Start one above index of known key and loop until the end
+        for (size_t j = known_key_index + 1; j < ring.size(); ++j)
+        {
+            BITCOIN_ASSERT(j < out.s[i].size());
```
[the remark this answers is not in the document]
narodnik: it should not be possible, if so it's a program error

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    out.e = sha256_hash(e0_data);
+
+    // ---------------------------------------------------------------------
+    // Step 2: join up each ring at the index where we know the secret key
+    // ---------------------------------------------------------------------
+    for (size_t i = 0; i < rings.size(); ++i)
+    {
+        BITCOIN_ASSERT(i < known_key_indexes.size());
+        BITCOIN_ASSERT(i < k.size());
+        BITCOIN_ASSERT(i < out.s.size());
```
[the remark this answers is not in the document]
narodnik: moved them out of the loop, and they should not be possible

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+
+        // Current ring and index of known key
+        const auto& ring = rings[i];
+        const size_t known_key_index = known_key_indexes[i];
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
narodnik: done

src/math/ring_signature.cpp
narodnik: moved it out of the loop

src/math/ring_signature.cpp
narodnik: it should not be possible

src/math/ring_signature.cpp
narodnik: it should not be possible

src/math/ring_signature.cpp
narodnik: moved it outside loop

src/math/ring_signature.cpp
narodnik: moved it outside loop

src/math/ring_signature.cpp
narodnik: removed

src/math/ring_signature.cpp
narodnik: Yes, done

src/math/ring_signature.cpp
narodnik: Yes done

src/math/ring_signature.cpp
narodnik: I cannot, it results in a compile error

src/math/ring_signature.cpp
narodnik: done

src/math/ring_signature.cpp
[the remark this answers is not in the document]
evoskuil: It doesn’t appear that iteration is necessary to compute the size. Are the values variable-sized? In any case the iteration would be much cheaper than the iterations that result from reapeatedly reallocating the vector and moving its data into the new contiguous memory.

src/math/ring_signature.cpp
narodnik: ok done

src/math/ring_signature.cpp
narodnik: ok done

test/math/ring_signature.cpp
narodnik: This is what the secp256k1-zkp unit test does. i just copied it.

test/math/ring_signature.cpp
narodnik: done

test/math/ring_signature.cpp
evoskuil: I don’t expect them to follow our practices, but we don’t want unit tests that change on every execution. The fauilures would be untraceable. Generation of one random but fixed vector is sufficient if randomness is sufficient.

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
evoskuil: Remember to remove const from the data slice parameter. For reasons described elsewhere this is avoided because the parameter is always constructed (ie passed by copy) and as such the type is not made safer by const, making it superfluous (unnecessary verbosity).

FWIW we always use const where possible in variable (vs. parameter) declarations because it declares that the intent is the variable not change in the scope, which can help avoid subtle mistakes (ie provides some type safety vale), but as a local variable is easily changed as necessary. Overconstraining a parameter (ie with const) has the opposite effect as it does not provide any interface safety but constrains internal implementation (as we should not modify interfaces once they are published). The same logic is applied to members, always const where possible. The one annoyance is that const members don’t work when implementing copy assignment (even though conceptually you are constructing a new instance).

src/math/ring_signature.cpp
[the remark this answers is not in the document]
evoskuil: The assertions are also error suppression. The error must be propagated to the caller.

src/math/ring_signature.cpp
```diff
@@ -203,34 +191,41 @@ ec_secret calculate_s(const ec_secret& k, const ec_secret& e,
...
+
     // Check we know a secret key in each ring
-    if (!all_rings_nonzero_size(known_keys_by_ring))
+    auto non_empty_ring = [](const point_list& ring)
+    {
+        return !ring.empty();
+    };
+    const bool all_rings_nonzero_size = std::all_of(
+        known_keys_by_ring.begin(), known_keys_by_ring.end(), non_empty_ring);
+
```
evoskuil: Suggest avoiding the double negative for readability and instead use “any ring zero size”.

include/bitcoin/bitcoin/math/ring_signature.hpp
narodnik: Ideally I don't want people modifying the e and s values themselves. In that sense, they are internal values, however there is no defined serialization format. They are abstract variables with no given names in the paper.

narodnik: R_i_j, e_0 (or e), e_i_j, s_i_j, k_i <- these are the values used

evoskuil: Okay, if the ring_sigature object has no public serialization it’s internals should remain hidden, and it should not be publicly constructible. To do this while maintaining essentially the same interface you can place the functions in a static class, define the ring_signature as a class in that context, give it a protected constructor only and expose its property methods and construction to only methods of that class.

narodnik: There's no defined serialization format, but still it needs to be serialized. So I'd need to access the internals in my code for saving the signature data.

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
narodnik: Thanks for the explanation, I can see that all the data members and variables are declared const.  Still the const is maybe desirable for readability and enforcement of constness if there's future changes. Readability shows that this is an object whose data is not meant to change. I find that a really useful indicator in code.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,363 @@
...
+    secret_keys_map keys;
+    for (const auto& secret: secrets)
+    {
+        // Compute public key
+        ec_compressed public_key;
+        bool rc = secret_to_public(public_key, secret);
+        BITCOIN_ASSERT(rc);
+        // Insert into map
+        keys.insert({public_key, secret});
+    }
```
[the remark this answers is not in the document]
narodnik: done

test/math/ring_signature.cpp
narodnik: OK, going to generate should standard test cases based off their unit test, translating the values to our code. Here is the code I'm referring to https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/rangeproof/tests_impl.h#L341

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,94 @@
...
+ */
+#ifndef LIBBITCOIN_RING_SIGNATURE_HPP
+#define LIBBITCOIN_RING_SIGNATURE_HPP
+
+#include <bitcoin/bitcoin/math/elliptic_curve.hpp>
+
+namespace libbitcoin {
+
+/**
+ * key_rings is a collection of public key rings.
```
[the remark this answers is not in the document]
narodnik: I corrected that line. I'm not sure with that would entail. Basically each ring of public keys represents 1 of N... A or B or C, and the rings together represent N of N, or (A or B or C) and (X or Y or Z)... where each ring can be a variable size.

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * That is given three rings [{A, B, C}, {D, E, F}, {X, Y}], then to
+ * create a valid signature, we must use a private key from each of those
+ * sets.
+ * For example A and E and X. We can summarize this operation as:
+ * (A or B or C) and (D or E or F) and (X or Y)
+ */
+typedef std::vector<point_list> key_rings;
+
+/**
+ * List of secret keys used for signing. There must be one valid secret
```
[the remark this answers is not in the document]
narodnik: ec_secret is used as an ec_scalar in this code. s has nothing to do with the signing keys. They are just numbers.

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,94 @@
...
+ * For example given a ring of [{A, B, C}, {D, E, F}, {X, Y}] then we
+ * must have a set of keys that satisfies this constraint:
+ * (A or B or C) and (D or E or F) and (X or Y)
+ */
+typedef std::vector<ec_secret> secret_list;
+
+/**
+ * A borromean ring signature.
+ *
+ * theta = {e_0, s_i_j : 0 <= i <= n, 0 <= j <= m_i}
```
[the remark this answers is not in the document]
narodnik: done

src/math/ring_signature.cpp
narodnik: I need to check the paper, but that's what the secp256k1-zkp code is doing too https://github.com/ElementsProject/secp256k1-zkp/blob/secp256k1-zkp/src/modules/rangeproof/borromean_impl.h#L81

include/bitcoin/bitcoin/math/ring_signature.hpp
evoskuil: This means we would be defining a public serialization. In that case if the values s and e are not publicly meaningful they shouldn’t be exposed. Instead declare public construction from a byte array|chunk and provide from_data/to_data methods consistent with other serializable wallet/chain/config types and make e and s private internals. Basically if these properties aren’t publicly meaningful they shouldn’t be public. In fact it seems more reasonable that the signature just be a byte array or chunk, just like the existing ec signature, and perform the parsing internal to the one function that needs/understands it.

include/bitcoin/bitcoin/utility/serializer.hpp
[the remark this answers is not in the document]
evoskuil: It’s not incorrect to use it, but it is superfluous apart from providing the hint that the value won’t be changed. However if the caller requires a variable not be changed a const variable can be passed. This stylistic preference is consistently applied to parameters passed by copy (including all data_slice) so that consistency should be maintained here if only as a matter of consistency.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,361 @@
...
+    BITCOIN_ASSERT(rc);
+    return result;
+}
+
+ec_secret calculate_s(const ec_secret& k, const ec_secret& e,
+    const ec_secret& secret)
+
+{
+    // s = k - e x
+    ec_secret result = e;
```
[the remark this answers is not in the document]
narodnik: done

include/bitcoin/bitcoin/math/ring_signature.hpp
```diff
@@ -0,0 +1,94 @@
...
+    s_values_type s;
+};
+
+/**
+ * Create a borromean ring signature.
+ * There must exist a valid signing key for each ring of public keys.
+ * @param[in]  out      The new signature.
+ * @param[in]  secrets  Secret signing keys. There should be at least one key
+ *                      from each ring.
+ * @param[in]  rings    The rings each with N_i public keys.
```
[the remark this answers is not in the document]
narodnik: done

include/bitcoin/bitcoin/math/ring_signature.hpp
narodnik: I can invent a data serialization format in the format [ EC_SECRET ][ N:4 ][ M_0:4 ][ EC_SECRET ]...[ EC_SECRET ][ M_1:4 ]...[ M_(N - 1):4 ][ EC_SECRET ]...[ EC_SECRET ] which would be 32 + 4 + 4 * N + (M_0 * 32 + M_1 * 32 + ... + M_(N - 1) * 32) bytes long. How does that sound?

src/math/ring_signature.cpp
narodnik: Yes, done.

src/math/ring_signature.cpp
narodnik: Maybe I'll just accept these parameters and for now let the user randomly generate them. Later we can add a wrapper which accepts a random seed.

include/bitcoin/bitcoin/math/ring_signature.hpp
evoskuil: This doesn’t look like a signature (ie contains secrets). It would seem all you need is [e][s].

evoskuil: The serialization doesn’t have to expose publicly its encoding, it can be treated as opaque. Nobody using libbitcoin needs to be able to externally encode/decode the wire/store serialization of any object.

src/math/ring_signature.cpp
evoskuil: That just puts the complexity on the user. The library logically only needs the seed, so publicly exposing its internals to avoid that is suboptimal from an interface standpoint. BIP32 requires that you handle the potential invalidity of a generated key by skipping to the next value. That can be done here internally, although it does present a halting problem. So it’s best just to return a failure in the case where any one fails just fail the function. The chance of it is very low, but needs to be handled.

thecodefactory:
This build failure is because master's install.sh isn't updated to version5 of secp256k1 (it's still at version4; libbitcoin version3 install.sh is correctly updated to secp256k1 version5).  @narodnik if you swap that out in the install.sh file and push it, travis-ci should just rebuild everything properly.  

Reference: https://github.com/libbitcoin/libbitcoin/blob/master/install.sh#L759

evoskuil:
Not sure how master fell behind. This requires an update to libbitcoin-build metadata and a regeneration for all repos.

src/math/ring_signature.cpp
```diff
@@ -0,0 +1,399 @@
...
+}
+
+bool calculate_e0(ring_signature& out, const key_rings& rings,
+    const hash_digest& digest, const secret_list& salts,
+    const index_list& known_key_indexes)
+{
+    data_chunk e0_data;
+    e0_data.reserve(ec_compressed_size * rings.size() + hash_size);
+
+    BITCOIN_ASSERT(known_key_indexes.size() == rings.size());
```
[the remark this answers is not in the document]
evoskuil: The treatment in terms of the choice between assertion and logical condition should be entirely determined by the context. If the methods is private and all calling methods have guarded against the condition an assertion is sufficient. All assertions should be entirely unnecessary and exist only to test what are believed to be *impossible* conditions. If the failure condition is exposed by the public/protected interface then an assertion is not allowed. I haven't looked to see which is the case here, but it should be unambiguous.

evoskuil:
Still some nits but will clean up after.

evoskuil:
Still some nits but will clean up after.

PR #949, opened 2018-05-11
Remove dead code about ec_public::version_

GingerMoon:
Remove dead code about ec_public::version_
[https://github.com/libbitcoin/libbitcoin/issues/946#issuecomment-388096665](url)


evoskuil:
Thanks @GingerMoon! Could you please squash these four commits to one and also backport this to version3?

Issue #788, opened 2017-05-08
[linux] log startup does not display local date/times.


evoskuil:
Should look like this:
```
09:22:30.974477 WARNING [node] ================= startup 05/08/17 02:22:30 ==================
09:22:30.985512 ERROR [node] ================= startup 05/08/17 02:22:30 ==================
09:22:30.995566 FATAL [node] ================= startup 05/08/17 02:22:30 ==================
```

evoskuil:
Is this resolved?

thecodefactory:
Yes (v3/master)!

PR #951, opened 2018-05-18
[linux] log startup does not display local date/times. #788

rockresu:
Changed buffer size to 25, to include the terminating null-character.


thecodefactory:
@rockresu Thanks, this does appear to resolve the issue!

PR #953, opened 2018-05-23
assertion fails for top=11, fixed

fpelliccioni:


thecodefactory:
@fpelliccioni Would you mind adding a test for this failure/correction?

evoskuil:
It doesn’t appear to me that this is correct. The last block will be picked up by the genesis block addition (which is accounted for). Where does the assertion arise?

fpelliccioni:
In the original code when you execute:

`locator_heights(11);`

the following assertion is violated:

`BITCOIN_ASSERT(heights.size() <= reservation);`

evoskuil:
I’m guessing that results from the unconditional inclusion of the genesis block entry in the case where it has already been included. That would only be the case when there is exactly one remaining, making the fix correct. Prob a comment to the effect that we don’t want to double count the genesis block would be helpful.

PR #965, opened 2018-06-06
Parametrize timespan constants.

toxeus:
The constants are parametrized such that libbitcoin can be used by various altcoins by solely editing the config file.

Further constants will be parametrized in upcoming pull requests.

As soon as this patch is in a satisfactory state I'll fix the unit tests and will open PRs in other libbitcoin repos to propagate these changes. Please comment.

### DIFF
--- a/examples/main.cpp
+++ b/examples/main.cpp
@@ -45,7 +45,7 @@ int bc::main(int argc, char* argv[])
 #endif
 
     // Extracting Satoshi's words from genesis block.
-    const auto block = bc::chain::block::genesis_mainnet();
+    const auto block = bc::chain::block::genesis_mainnet(bc::settings());
     const auto& coinbase = block.transactions().front();
     const auto& input = coinbase.inputs().front();
 	BITCOIN_ASSERT_MSG(input.script().size() > 2u, "unexpected genesis");
--- a/include/bitcoin/bitcoin/chain/chain_state.hpp
+++ b/include/bitcoin/bitcoin/chain/chain_state.hpp
@@ -23,6 +23,7 @@
 #include <cstdint>
 #include <memory>
 #include <deque>
+#include <bitcoin/bitcoin/settings.hpp>
 #include <bitcoin/bitcoin/config/checkpoint.hpp>
 #include <bitcoin/bitcoin/constants.hpp>
 #include <bitcoin/bitcoin/define.hpp>
@@ -125,23 +126,25 @@ public:
 
     /// Checkpoints must be ordered by height with greatest at back.
     static map get_map(size_t height, const checkpoints& checkpoints,
-        uint32_t forks);
+        uint32_t forks, size_t retargeting_interval);
 
     static uint32_t signal_version(uint32_t forks);
 
     /// Create pool state from top chain top block state.
-    chain_state(const chain_state& top);
+    chain_state(const chain_state& top, const settings& settings);
 
     /// Create block state from tx pool chain state of same height.
-    chain_state(const chain_state& pool, const chain::block& block);
+    chain_state(const chain_state& pool, const chain::block& block,
+        const settings& settings);
 
     /// Create header state from header pool chain state of parent block.
-    chain_state(const chain_state& parent, const chain::header& header);
+    chain_state(const chain_state& parent, const chain::header& header,
+        const settings& settings);
 
     /// Checkpoints must be ordered by height with greatest at back.
     /// Forks and checkpoints must match those provided for map creation.
     chain_state(data&& values, const checkpoints& checkpoints, uint32_t forks,
-        uint32_t stale_seconds);
+        uint32_t stale_seconds, const settings& settings);
 
     /// Properties.
     const hash_digest& hash() const;
@@ -182,29 +185,40 @@ protected:
 
     static activations activation(const data& values, uint32_t forks);
     static uint32_t median_time_past(const data& values, uint32_t forks);
-    static uint32_t work_required(const data& values, uint32_t forks);
+    static uint32_t work_required(const data& values, uint32_t forks,
+        const settings& settings);
 
 private:
-    static size_t bits_count(size_t height, uint32_t forks);
+    static size_t bits_count(size_t height, uint32_t forks,
+        size_t retargeting_interval);
     static size_t version_count(size_t height, uint32_t forks);
     static size_t timestamp_count(size_t height, uint32_t forks);
-    static size_t retarget_height(size_t height, uint32_t forks);
+    static size_t retarget_height(size_t height, uint32_t forks,
+        size_t retargeting_interval);
     static size_t bip9_bit0_height(size_t height, uint32_t forks);
     static size_t bip9_bit1_height(size_t height, uint32_t forks);
 
-    static data to_pool(const chain_state& top);
+    static data to_pool(const chain_state& top, const settings& settings);
     static data to_block(const chain_state& pool, const block& block);
-    static data to_header(const chain_state& parent, const header& header);
+    data to_header(const chain_state& parent, const header& header,
+        const settings& settings);
 
-    static uint32_t work_required_retarget(const data& values);
-    static uint32_t retarget_timespan(const chain_state::data& values);
+    static uint32_t work_required_retarget(const data& values,
+        uint32_t retarget_proof_of_work_limit, uint32_t min_timespan,
+        uint32_t max_timespan, uint32_t target_timespan_seconds);
+    static uint32_t retarget_timespan(const chain_state::data& values,
+        uint32_t min_timespan, uint32_t max_timespan);
 
     // easy blocks
-    static uint32_t easy_work_required(const data& values);
-    static uint32_t easy_time_limit(const chain_state::data& values);
-    static bool is_retarget_or_non_limit(size_t height, uint32_t bits);
-    static bool is_retarget_height(size_t height);
-    static size_t retarget_distance(size_t height);
+    static uint32_t easy_work_required(const data& values,
+        size_t retargeting_interval, uint32_t retarget_proof_of_work_limit,
+        uint32_t easy_spacing_seconds);
+    static uint32_t easy_time_limit(const chain_state::data& values,
+        int64_t spacing);
+    static bool is_retarget_or_non_limit(size_t height, uint32_t bits,
+        size_t retargeting_interval, uint32_t retarget_proof_of_work_limit);
+    static bool is_retarget_height(size_t height, size_t retargeting_interval);
+    static size_t retarget_distance(size_t height, size_t retargeting_interval);
 
     // This is retained as an optimization for other constructions.
     // A similar height clone can be partially computed, reducing query cost.
--- a/include/bitcoin/bitcoin/settings.hpp
+++ b/include/bitcoin/bitcoin/settings.hpp
@@ -1,0 +1,55 @@
+/**
+ * Copyright (c) 2011-2018 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_SETTINGS_HPP
+#define LIBBITCOIN_SETTINGS_HPP
+
+#include <bitcoin/bitcoin/define.hpp>
+
+#include <cstdint>
+#include <cstddef>
+
+namespace libbitcoin {
+
+/// Common database configuration settings, properties not thread safe.
+class BC_API settings
+{
+public:
+    settings();
+
+    uint32_t work_limit(bool retarget=true) const;
+
+    uint32_t retargeting_factor;
+    uint32_t target_spacing_seconds;
+    uint32_t easy_spacing_seconds;
+    uint32_t timestamp_future_seconds;
+    uint32_t target_timespan_seconds;
+    uint32_t retarget_proof_of_work_limit;
+    uint32_t no_retarget_proof_of_work_limit;
+
+    // The upper and lower bounds for the retargeting timespan.
+    uint32_t min_timespan;
+    uint32_t max_timespan;
+
+    // The target number of blocks for 2 weeks of work (2016 blocks).
+    size_t retargeting_interval;
+};
+
+} // namespace libbitcoin
+
+#endif
--- a/src/chain/header.cpp
+++ b/src/chain/header.cpp
@@ -40,8 +40,8 @@ using wall_clock = std::chrono::system_clock;
 // Constructors.
 //-----------------------------------------------------------------------------
 
-header::header()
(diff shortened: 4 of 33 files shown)


src/chain/header.cpp
```diff
@@ -41,7 +41,7 @@ using wall_clock = std::chrono::system_clock;
 //-----------------------------------------------------------------------------
 
 header::header()
-  : header(0, null_hash, null_hash, 0, 0, 0)
+  : header(0, null_hash, null_hash, 0, 0, 0, settings{})
```
toxeus: `header::header()` is used by the factory functions below. I added this dummy value such that we can have the discussion here based on the code. I'd fix this by adding a `settings` parameter to this constructor and the factory functions. @evoskuil what do you think?

include/bitcoin/bitcoin/chain/chain_state.hpp
```diff
@@ -124,24 +125,26 @@ class BC_API chain_state
     };
 
     /// Checkpoints must be ordered by height with greatest at back.
-    static map get_map(size_t height, const checkpoints& checkpoints,
+    map get_map(size_t height, const checkpoints& checkpoints,
```
evoskuil: Keep this static and parameterize necessary settings (or settings object).

include/bitcoin/bitcoin/chain/chain_state.hpp
```diff
@@ -181,29 +184,29 @@ class BC_API chain_state
 
     static activations activation(const data& values, uint32_t forks);
     static uint32_t median_time_past(const data& values, uint32_t forks);
-    static uint32_t work_required(const data& values, uint32_t forks);
+    uint32_t work_required(const data& values, uint32_t forks);
```
evoskuil: Keep this static and parameterize necessary settings (or settings object).

src/chain/header.cpp
evoskuil: Yes, require the parameter until all other repos have been updated, so that you do not end up missing the necessary settings passage. But at that point you can also add a defaulted construction.

src/chain/header.cpp
```diff
@@ -65,32 +66,35 @@ header::header(const header& other)
...
     nonce_(other.nonce_),
-    metadata(other.metadata)
+    metadata(other.metadata),
+    settings_(other.settings_)
 {
 }
 
 header::header(uint32_t version, hash_digest&& previous_block_hash,
-    hash_digest&& merkle, uint32_t timestamp, uint32_t bits, uint32_t nonce)
+    hash_digest&& merkle, uint32_t timestamp, uint32_t bits, uint32_t nonce, const settings& settings)
```
evoskuil: Style: honor the 80 character line limit (more above and below).

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,62 @@
...
+    uint32_t min_timespan;
+    uint32_t max_timespan;
+
+    // The target number of blocks for 2 weeks of work (2016 blocks).
+    size_t retargeting_interval;
+
+    uint32_t work_limit(bool retarget=true) const
+    {
+        return retarget ? retarget_proof_of_work_limit : no_retarget_proof_of_work_limit;
+    }
```
evoskuil: Move implementation to cpp.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,62 @@
...
+    uint32_t no_retarget_proof_of_work_limit;
+
+    // The upper and lower bounds for the retargeting timespan.
+    uint32_t min_timespan;
+    uint32_t max_timespan;
+
+    // The target number of blocks for 2 weeks of work (2016 blocks).
+    size_t retargeting_interval;
+
+    uint32_t work_limit(bool retarget=true) const
```
evoskuil: Style: move declaration above fields.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,62 @@
...
+    {
+        return retarget ? retarget_proof_of_work_limit : no_retarget_proof_of_work_limit;
+    }
+
```
evoskuil: Style: remove blank line.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,62 @@
...
+#include <cstdint>
+#include <cstddef>
+
+namespace libbitcoin {
+
+/// Common database configuration settings, properties not thread safe.
+class BC_API settings
+{
+
+private:
```
evoskuil: Style: private after public: https://github.com/libbitcoin/libbitcoin/wiki/Style-Guide#hpp-format

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,62 @@
...
+
+#include <cstdint>
+#include <cstddef>
+
+namespace libbitcoin {
+
+/// Common database configuration settings, properties not thread safe.
+class BC_API settings
+{
+
```
evoskuil: Style: remove blank line: https://github.com/libbitcoin/libbitcoin/wiki/Style-Guide#hpp-format

src/chain/header.cpp
toxeus: Alright, I have propagated the `settings` as a parameter wherever needed. This blew up the PR to some extend. Now, I know what you meant by this refactoring being "pervasive" :) Pushing new version now.

toxeus:
I'm still postponing to fix the tests until I know that the main changes are ok.

examples/main.cpp
```diff
@@ -45,7 +45,8 @@ int bc::main(int argc, char* argv[])
 #endif
 
     // Extracting Satoshi's words from genesis block.
-    const auto block = bc::chain::block::genesis_mainnet();
+    const auto block =
+        bc::chain::block::genesis_mainnet(libbitcoin::settings());
```
evoskuil: use `bc::settings()`

evoskuil: The genesis block needs to itself become a setting, meaning that `genesis_mainnet` will no longer exist. The genesis block will be deserialized from the base16-encoded setting stream.

include/bitcoin/bitcoin/chain/chain_state.hpp
```diff
@@ -181,29 +184,39 @@ class BC_API chain_state
...
+    static data to_pool(const chain_state& top, const settings& settings);
     static data to_block(const chain_state& pool, const block& block);
-    static data to_header(const chain_state& parent, const header& header);
+    data to_header(const chain_state& parent, const header& header,
+        const settings& settings);
 
-    static uint32_t work_required_retarget(const data& values);
-    static uint32_t retarget_timespan(const chain_state::data& values);
+    static uint32_t work_required_retarget(const data& values,
+        const settings& settings);
```
evoskuil: Inconsistency here, pass `retarget_proof_of_work_limit`.

src/message/compact_block.cpp
```diff
@@ -99,31 +99,33 @@ bool compact_block::is_valid() const
...
+bool compact_block::from_data(uint32_t version, std::istream& stream,
+    const settings& settings)
 {
     istream_reader source(stream);
-    return from_data(version, source);
+    return from_data(version, source, settings);
 }
 
-bool compact_block::from_data(uint32_t version, reader& source)
+bool compact_block::from_data(uint32_t version, reader& source, const settings& settings)
```
evoskuil: Style: 80 char line length.

evoskuil:
See warnings:
```
  CXX      src/src_libbitcoin_la-settings.lo
src/settings.cpp:33:44: warning: field 'retargeting_factor_' is uninitialized when used here [-Wuninitialized]
    min_timespan(target_timespan_seconds / retargeting_factor_),
                                           ^
src/settings.cpp:34:44: warning: field 'retargeting_factor_' is uninitialized when used here [-Wuninitialized]
    max_timespan(target_timespan_seconds * retargeting_factor_),
                                           ^
src/settings.cpp:35:52: warning: field 'target_spacing_seconds_' is uninitialized when used here [-Wuninitialized]
    retargeting_interval(target_timespan_seconds / target_spacing_seconds_)
                                                   ^
3 warnings generated.
  CXX      src/chain/src_libbitcoin_la-block.lo
  CXX      src/chain/src_libbitcoin_la-chain_state.lo
src/chain/chain_state.cpp:124:29: warning: comparison of unsigned expression >= 0 is always true [-Wtautological-compare]
         (regtest && height >= regtest_bip34_freeze));
                     ~~~~~~ ^  ~~~~~~~~~~~~~~~~~~~~
1 warning generated.
```

examples/main.cpp
toxeus: Yes, the genesis block is on my todo list. I prefer to approach the migration of the constants iteratively. Once this is merged I'd open a new PR where the genesis block becomes a setting.

toxeus:
@evoskuil I was wondering why I didn't see those warnings. Turns out that they are [enabled for C only](https://github.com/libbitcoin/libbitcoin/blob/52258bf331bb24e3d8e0616cbc1d0fabaad0053a/configure.ac#L307-L332), i.e. they're assigned to `CFLAGS` as opposed to `CXXFLAGS` which is used above and below the snipped I highlighted in the link. Also, `CFLAGS` is not eventually appended to `CXXFLAGS`. Now, I'm wondering why you *did* see the warnings!

include/bitcoin/bitcoin/chain/chain_state.hpp
toxeus: If you check the implementation then you can see that actually four values from `settings` are used in `work_required_retarget()`:
* `settings.retarget_proof_of_work_limit`
* `settings.min_timespan`
* `settings.max_timespan`
* `settings.target_timespan_seconds`.
I'm happy to pass them individually but my judgement was that they would bloat the function signature. Your call.

examples/main.cpp
toxeus: done

src/message/compact_block.cpp
toxeus: done

toxeus:
> See the Travis builds for the warnings.

@evoskuil that was not my point. I've fixed the warnings in the second commit already. I just wanted to raise awareness that `configure.ac` is currently not set up correctly. Maybe @pmienk can confirm that?

toxeus:
gotcha :+1:

evoskuil:
@toxeus The warnings (in fact errors) were the result of field ordering in the header file. Fields are initialized in the header order, not in the order listed in the constructor initializers. I always try to keep these orders the same to avoid this type of failure. The errors were resolved by moving the initialization of these values into the constructor body.

There should not be private fields in the settings class. Making these public and matching the declaration order to the initialization order will resolve the issue properly.

toxeus:
@evoskuil ok, I'm happy to fix the initialization issue the way you've suggested instead of moving it to the constructor body. But what about the build systemt's bug where it doesn't enable warnings for cpp files? Do you prefer me to open an issue for that?

evoskuil:
@toxeus I’m not seeing the issue you are describing. These warnings are in a cpp file. But if you want to open an issue with explanation the libbitcoin-build repository is the best place to do it. Thanks!

toxeus:
@evoskuil I gave you a curl command to see the issue I'm describing. I'll open an issue in the repo you've suggested. Thanks!

evoskuil:
@toxeus I pointed out above the the warnings do appear in the Travis builds. The specific builds from the curl script did not have the warnings only because you inadvertently resolved them.

toxeus:
@evoskuil the curl script fetches the logs for the commit where the warnings should have been present. Visit the travis build link and then click on commit.

toxeus:
@evoskuil Strange. Maybe travis overwrites the logs.

evoskuil:
Just look back at previous builds, the warnings are there in the logs.

toxeus:
@evoskuil wrote:
> Just look back at previous builds, the warnings are there in the logs.

Actually, it's not in the previous build but in the [next](https://travis-ci.org/libbitcoin/libbitcoin/builds/389539885) one. 

Now, I remember how this all came about!
I originally had an initialization order that *did* match the declaration order in the header files in my first version because I also always try to make them match (that's why I couldn't find the warning in the first travis build). Then, you asked me to adhere to the libbitcoin style and move the private members to the end of the header file. I've blindly implemented that suggestion which obviously introduced a mismatch in order between declaration and initialization. And I didn't see the corresponding warnings because I'm using gcc for development and the build system doesn't enable the warnings for gcc. See opened issue [here](https://github.com/libbitcoin/libbitcoin-build/issues/163). So, it turns out that we were both right about the presence of the warnings! :wink:

evoskuil:
Great, thanks for tracking it down. I was only looking at the clang builds. I'll check out the -build issue.

include/bitcoin/bitcoin/chain/chain_state.hpp
evoskuil: Yeah, these aren't always obvious calls. I've tried for a consistent approach whereby classes are constructed with the the settings object given sufficient use, but functions (including static members) are parameterized so as to avoid a dependency on the setting object.

evoskuil:
Time to update the tests and get it passing.

toxeus:
@evoskuil I've fixed the test. Will start work on the PRs for the other repos.

src/settings.cpp
```diff
@@ -0,0 +1,45 @@
...
+    target_timespan_seconds(2 * 7 * 24 * 60 * 60),
+    retarget_proof_of_work_limit(0x1d00ffff),
+    no_retarget_proof_of_work_limit(0x207fffff),
+    min_timespan(target_timespan_seconds / retargeting_factor),
+    max_timespan(target_timespan_seconds * retargeting_factor),
+    retargeting_interval(target_timespan_seconds / target_spacing_seconds)
+{
+}
+
+uint32_t settings::work_limit(bool retarget) const
```
evoskuil: We should move this function out of settings and into chain::header (see chain::header::proof). Prob should do that in an independent commit.

test/chain/block.cpp
```diff
@@ -262,30 +266,30 @@ BOOST_AUTO_TEST_CASE(block__from_data__insufficient_transaction_bytes__failure)
...
-    chain::block instance;
+    chain::block instance(settings{});
     BOOST_REQUIRE(!instance.from_data(data));
     BOOST_REQUIRE(!instance.is_valid());
 }
 
 BOOST_AUTO_TEST_CASE(block__genesis__mainnet__valid_structure)
 {
-    const auto genesis = bc::chain::block::genesis_mainnet();
+    const auto genesis = bc::chain::block::genesis_mainnet(settings());
```
evoskuil: Use consistent style (settings() and settings{}).

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -0,0 +1,55 @@
+/**
```
evoskuil: Need to add tests for settings class (see other repos).

test/chain/block.cpp
toxeus: My style is very consistent. I always use `settings()` except for cases where the vexing parse occurs.

src/settings.cpp
toxeus: Independent commit as in independent PR?

toxeus:
Rebased and fixed conflicts.

src/settings.cpp
toxeus: @evoskuil I've been now looking into moving this function into chain::header as you've suggested. The problem is that it is also used in [chain::chain_state](https://github.com/libbitcoin/libbitcoin/blob/2e1e6792736e73f8a216ea09038084c01f22972a/src/chain/chain_state.cpp#L604) where there's no access to a `header` object. However, there's a comment which states that `bits.self` is unused. If this assignment can be removed then moving `work_limit()` becomes trivial.

evoskuil: Yes, I you can set it to zero in that case. Self has no purpose in the context of a pool chain state and when the pool state is upgraded to a block the blocks bits are populated.

PR #967, opened 2018-06-07
Update boost version to 1.62.0

thecodefactory:


thecodefactory:
This was done via libbitcoin-build generation.  If all builds well, I'll update the other repos and libbitcoin-build after.

evoskuil:
I recommend making a pull request in each repo before merging any, since the builds are independent and this may surface issues not found in the system repo.

pmienk:
Are we abandoning support of versions below 1.62?  Looks like you're just rerunning the generator, so not much to review codewise assuming no changes to anything but xml in libbitcoin-build.  As long as 1.62 is the minimum value provided by travis across the test matrix we'll be fine.  It's certainly been a moving target before.

evoskuil:
That’s the implication. There are boost breaks in higher versions of gcc.

thecodefactory:
@evoskuil Yes, that's the plan on issuing the PR/ordering.  @pmienk Yes, that was the idea.  Any objections?

pmienk:
Fine by me, raised this last year related to logging.  We might be able to shave some of the duplication now.  It looks like travis is providing at or above 1.62 on boost-latest, so we're all good.

PR #969, opened 2018-06-10
Add ec_scalar and ec_point classes and update ring_signature.cpp

narodnik:
These classes simplify EC arithmetic operations and should be two-way compatible with ec_secret and ec_compressed:

```cpp
ec_compressed Q = ...;
ec_secret z = ...;
ec_secret e = ...;
const auto Z = Q * z + e * ec_point::G;
if (!Z)
    return false;
ec_compressed Z_converted = Z;
```

The compiler will automatically do the conversions implicitly back and forth.

If a single operation in a calculation fails, then the value will become invalid, and all subsequent operations on that value should fail. This means we can error check afterwards when doing a complicated calculation.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -91,6 +91,7 @@ src_libbitcoin_la_SOURCES = \
     src/machine/program.cpp \
     src/math/checksum.cpp \
     src/math/crypto.cpp \
+    src/math/ec_arithmetic.cpp \
     src/math/elliptic_curve.cpp \
     src/math/ring_signature.cpp \
     src/math/hash.cpp \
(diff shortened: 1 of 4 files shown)


evoskuil:
Thanks! Four big picture comments:

1) Please rebase on master vs. creating a merge commit (the code is 11 commits behind master and makes review of the specific changes very difficult).
2) Please split ec_arithmetic into ec_point and ec_scalar (style: one class per file, file name matches class).
3) Please make the two new classes bases for the corresponding ec_wallet classes (see previous discussion).
4) Please add unit test coverage for the new classes.

narodnik:
I also added a new constant in ec_scalar called zero, and removed ec_scalar::is_zero() since the user can easily do `x == ec_scalar::zero` now.

evoskuil:
Is there ever a case where a zero-valued scalar is considered valid? Put another way, under what condition would one check for zero value? If it is always invalid it can be used as the invalidity sentinel.

evoskuil:
Please see outstanding comments 1 & 3 (git rebase and wallet class basing).

narodnik:
Sometimes the scalar can be zero but valid. They are different things.

evoskuil:
Why is zero special then?

evoskuil:
@naridnik this needs to be rebased.

narodnik:
It isn't special. But the ring signature algorithm treats zero as an error value. However other code (pedersen commitments), zero is acceptable.

evoskuil:
Okay, thanks. Was just curious.

evoskuil:
See warning and error [here](https://ci.appveyor.com/project/evoskuil/libbitcoin/build/4.0.0.1395-master/job/mbm5rccpgdsy68ll).

evoskuil:
I'm working on a commit that resolves the above warnings and error, regenerates build artifacts for VS projects, fixes some style deviations, and properly integrates sub-classing by the two corresponding wallet classes.

PR #981, opened 2018-07-09
Fix MSVC unicode stream regression.

evoskuil:
@pmienk This was introduced in the warnings fix. It would only be caught in the MSVC test run.


pmienk:
Clearly a mistake on my part inferring the type.  At least now that appveyor is working again it could be more easily found.

PR #982, opened 2018-07-09
Fix MSVC builds, style.

evoskuil:


evoskuil:
@toxeus You should get familiar with execution of the libbitcoin-build generation script so that all platform builds are kept current. We don't presently have sufficient CI coverage to ensure the MSVC builds.

toxeus:
Alright :+1:

PR #987, opened 2018-07-11
Warning cleanups.

thecodefactory:


src/utility/png.cpp
```diff
@@ -80,7 +78,8 @@ bool png::write_png(std::istream& in, uint32_t size, uint32_t dots_per_inch,
         return false;
 
     istream_reader source(in);
-    auto version = source.read_4_bytes_little_endian();
+    // Skip version.
+    source.read_4_bytes_little_endian();
```
evoskuil: You should be able to use source.skip(4) here.

src/utility/png.cpp
```diff
@@ -95,9 +94,14 @@ bool png::write_png(std::istream& in, uint32_t size, uint32_t dots_per_inch,
         static constexpr int32_t bits_per_byte = 8;
         static constexpr uint8_t margin_value = 0xff;
 
-        // TODO: unguarded overflow conditions.
         const auto margin_size = margin * size;
+        if ((size == 0) || (margin_size / size) != margin)
```
evoskuil: I'm not confident these round trip checks are sufficient guards. The guard should require the domain limit.

PR #988, opened 2018-07-11
parametrize genesis block

toxeus:
The libbitcoin-build script didn't change anything so I'm assuming that I didn't break the VS build. I'm fixing the other repos now.


evoskuil:
The build script only updates build artifacts. It cannot verify the compile or test execution - that's up to the automated build platforms. So you could have broken a VS or other platform build, but at least the project files are correct.

toxeus:
My understanding was that the project files are the Makefiles of the VS universe. If Makefiles are outdated then the build can break. But of course it can break for other reasons too :wink:

evoskuil:
Right, so the script only ensures that the project and make files are generated for current sources, you can still break the VS build.

src/settings.cpp
```diff
@@ -36,6 +36,94 @@ settings::settings()
...
 }
 
+settings::settings(config::settings context)
+  : settings()
+{
+    switch (context)
+    {
+        case config::settings::mainnet:
+        {
+            encoded_genesis_block =
```
evoskuil: The settings pattern uses serializable classes to read settings to produce the corresponding object. I think the proper generalization here is actually a chain::block.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -48,6 +51,8 @@ class BC_API settings
 
     // The target number of blocks for 2 weeks of work (2016 blocks).
     size_t retargeting_interval;
+
+    std::string encoded_genesis_block;
```
evoskuil: `chain::block genesis_block` (more below)

src/settings.cpp
toxeus: This would make the `chain::block::genesis()` function obsolete. I'll remove it then.

evoskuil: It makes it redundant, but that’s fine. It can accept only the standard config enumeration and construct the well know instances, as it does now, and those can be used to simplify the standard settings init.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -31,6 +34,9 @@ class BC_API settings
 {
 public:
     settings();
+    settings(config::settings context);
+
+    void genesis(const std::string& encoded_genesis_block);
```
evoskuil: Should be private.

src/settings.cpp
```diff
@@ -32,10 +33,110 @@ settings::settings()
     no_retarget_proof_of_work_limit(0x207fffff),
     min_timespan(target_timespan_seconds / retargeting_factor),
     max_timespan(target_timespan_seconds * retargeting_factor),
-    retargeting_interval(target_timespan_seconds / target_spacing_seconds)
+    retargeting_interval(target_timespan_seconds / target_spacing_seconds),
+    genesis_block(*this)
 {
 }
 
+settings::settings(config::settings context)
```
evoskuil: The use of text deserialization internal to the settings instance is inconsistent. Setting and command line deserialization is implemented in `bc::config` classes, often encapsulating the primitive. Additionally, use of genesis(text) below is unnecessary. There is no reason to construct the object from a text string. The text stream is broken up into (sort of) logical values for readability. But this is much more clearly and safely accomplished using the `chain::block` constructor as it is intended. Comparing the result to a well-known text stream and block hash result is the job of two text cases.

src/chain/block.cpp
```diff
@@ -200,7 +140,7 @@ bool block::operator!=(const block& other) const
 //-----------------------------------------------------------------------------
 
 // static
-block block::factory(const data_chunk& data, const settings& settings,
+block block::factory(const data_chunk& data, const bc::settings& settings,
```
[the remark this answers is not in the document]
toxeus: It conflicts with `bc::config::settings`.

src/settings.cpp
toxeus: Will fix.

examples/main.cpp
```diff
@@ -45,7 +45,8 @@ int bc::main(int argc, char* argv[])
 #endif
 
     // Extracting Satoshi's words from genesis block.
-    const auto block = bc::chain::block::genesis_mainnet(bc::settings());
+    const auto block =
+        bc::settings(bc::config::settings::mainnet).genesis_block;
```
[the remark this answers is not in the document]
toxeus: Will fix.

include/bitcoin/bitcoin/settings.hpp
toxeus: This will go away as you suggested below.

src/settings.cpp
toxeus: I have removed the `genesis()` helper method. I will add the deserialization class in `bc::config` in an upcoming PR to prevent this one from growing too much.

evoskuil:
see: https://github.com/libbitcoin/libbitcoin/pull/997

PR #992, opened 2018-07-16
add config::block class.

toxeus:
The first intended use of this class is deserializing the genesis block.

### DIFF
--- a/test/config/block.cpp
+++ b/test/config/block.cpp
@@ -1,0 +1,118 @@
+/**
+ * Copyright (c) 2011-2018 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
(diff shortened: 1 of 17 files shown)


toxeus:
The AppVeyor build fails with
> C:\Program Files (x86)\MSBuild\Microsoft.Cpp\v4.0\V120\Microsoft.Cpp.Platform.targets(64,5): error MSB8020: The build tools for CTP_Nov2013 (Platform Toolset = 'CTP_Nov2013') cannot be found. To build using the CTP_Nov2013 build tools, please install CTP_Nov2013 build tools.  Alternatively, you may upgrade to the current Visual Studio tools by selecting the Project menu or right-click the solution, and then selecting "Upgrade Solution...". [c:\git\libbitcoin\builds\msvc\vs2013\libbitcoin\libbitcoin.vcxproj]

Doesn't look like it's related to this patch.

test/config/block.cpp
```diff
@@ -0,0 +1,118 @@
...
+    "04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73"
+    "ffffffff"
+    "01"
+    "00f2052a01000000"
+    "43"
+    "4104678afdb0fe5548271967f1a67130b7105cd6a828e03909a67962e0ea1f61deb649f6bc3f4cef38c4f35504e51ec112de5c384df7ba0b8d578a4c702b6bf11d5fac"
+    "00000000";
+
+
+const static auto genesis_block = block(encoded_genesis_block, settings);
```
evoskuil: style: `static const`

toxeus: Fixed.

evoskuil:
@pmienk is looking into the appveyor thing, it’s not related to your changes.

PR #995, opened 2018-07-17
Regenerate artifacts.

pmienk:


toxeus:
Thanks for the fix :+1:

PR #1004, opened 2018-08-08
Parametrize currency unit constants.

toxeus:


toxeus:
@evoskuil because this PR touches quite a lot of different classes I'd like to ask you for early feedback before I fix the tests and propagate the changes in the other repos.

src/machine/program.cpp
```diff
@@ -49,36 +48,38 @@ void program::reserve_stacks()
...
+    settings_(settings)
 {
     reserve_stacks();
 }
 
-program::program(const script& script)
+program::program(const script& script, const settings& settings)
   : script_(script),
-    transaction_(default_tx_),
+    transaction_(std::move(chain::transaction(settings))),
```
evoskuil: Move not useful here, it is already an r-value.

toxeus:
A `settings_` member variable was introduced in `chain::transaction` and hence the constructor needed to be adapted which caused a lot of different classes to propagate the `bc::settings` object. We might get away with far less of the code that just propagates `bc::settings` if we pass the `bc::settings` object to the method that actually needs it. I'll dig into it.

toxeus:
Turns out I was right. Now, the patch has shrunk quite a bit.

src/config/output.cpp
```diff
@@ -35,14 +35,14 @@ namespace config {
 
 using namespace boost::program_options;
 
-output::output()
+output::output(const bc::settings& settings)
```
toxeus: Losing this default constructor turns out to be a problem. In `bc::explorer::commands::tx-encode` this is used for `boost::program_options` which in turn uses `boost::lexical_cast` internally. `boost::lexical_cast` requires the target object to be default-constructible.

evoskuil:
The point I was making is that the tx in sighash computation can be copy constructed.

toxeus:
I see. Still, the patch will increase significantly compared to the current version. But I guess it's more consistent if the settings are passed via constructor.

toxeus:
I have updated the PR to the previous version but have incorporated your suggestion to copy construct the tx in the sighash computation. The issue with issue with `bc::config::output` not having a default constructor and therefore not being usable by `boost::program_options` is still present. I'm not sure what to do about it because the affected code in libbitcoin-explorer is generated.

src/machine/program.cpp
```diff
@@ -49,9 +48,9 @@ void program::reserve_stacks()
 // Constructors.
 //-----------------------------------------------------------------------------
 
-program::program()
+program::program(const settings& settings)
   : script_(default_script_),
-    transaction_(default_tx_),
+    transaction_(chain::transaction(settings)),
```
toxeus: Here, we're binding a temporary to a const reference which will result in a dangling reference after the ctor exits. I propose to change the type of `transaction_` to `std::shared_ptr` to fix this.

src/config/output.cpp
evoskuil: The config classes should all allow default construction. In this case it's acceptable to just remove the amount validation. In the case of header (and therefore block) it appears that we could eliminate the need for settings by parameterizing:
```cpp
    bool is_valid_timestamp() const;
    bool is_valid_proof_of_work(bool retarget) const;

    code check(bool retarget) const;
```

Though I'm not sure about the downstream impact. I suppose this brings us back to your previous suggestion.

src/machine/program.cpp
evoskuil: That would result in a copy of the transaction when the tx is passed. The issue would be resolved by eliminating the settings constructor.

evoskuil: Ha, that’s not actually a shared pointer, that’s a weak pointer.

evoskuil: I do believe eliminating the members for the chain (and config) classes would be better.

toxeus: OK. I have no idea though how do get rid of the member in config because it's used in `std::istream& operator>>(std::istream& input, output& argument)` so I cannot pass it in as an argument. The only thing that comes to mind is to remove this check
```cpp
    if (amount > argument.settings_.max_money())
    {
        BOOST_THROW_EXCEPTION(invalid_option_value(tuple));
    }
```

evoskuil: See previous comment. Eliminate this test.

toxeus: > I do believe eliminating the members for the chain (and config) classes would be better.

Done.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -36,6 +36,15 @@ class BC_API settings
...
     settings(config::settings context);
 
+    // Currency unit helper methods (uint64_t).
+    //--------------------------------------------------------------------------
+
+    uint64_t max_money_recursive(uint64_t money) const;
+    uint64_t bitcoin_to_satoshi(uint64_t bitcoin_uints=1) const;
+    uint64_t initial_block_subsidy_satoshi() const;
+    uint64_t subsidy_interval(bool retarget=true) const;
+    uint64_t max_money(bool retarget=true) const;
```
evoskuil: Logic should be kept out of the settings classes. I've only used this approach for simple data type conversions (e.g. integer to time). These functions can remain centralized or moved to class static helpers.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -88,6 +97,15 @@ class BC_API settings
...
     config::checkpoint bip9_bit1_active_checkpoint;
+
+    // Currency unit settings (uint64_t).
+    //--------------------------------------------------------------------------
+
+    uint64_t satoshi_per_bitcoin;
+    uint64_t initial_block_subsidy_bitcoin;
+    uint64_t recursive_money;
+    uint64_t retarget_subsidy_interval;
+    uint64_t no_retarget_subsidy_interval;
```
evoskuil: `retarget_subsidy_interval` and `no_retarget_subsidy_interval` should be combined into `subsidy_interval`. This will eliminate need to pass both the `fork.retarget' setting and this value in cases where the interval is the only condition, which should simplify a bunch of code.

src/chain/transaction.cpp
```diff
@@ -1092,15 +1093,16 @@ code transaction::connect_input(const chain_state& state,
 //-----------------------------------------------------------------------------
 
 // These checks are self-contained; blockchain (and so version) independent.
-code transaction::check(bool transaction_pool, bool retarget) const
+code transaction::check(const settings& settings, bool transaction_pool,
+    bool retarget) const
```
evoskuil: The `retarget` parameter is redundant with the `settings` parameter.

src/chain/transaction.cpp
```diff
@@ -1092,15 +1093,16 @@ code transaction::connect_input(const chain_state& state,
...
 {
     if (inputs_.empty() || outputs_.empty())
         return error::empty_transaction;
 
     else if (is_null_non_coinbase())
         return error::previous_output_null;
 
-    else if (total_output_value() > max_money(retarget))
+    else if (total_output_value() > settings.max_money(retarget))
```
evoskuil: There will be no need to pass `retarget` here.

include/bitcoin/bitcoin/chain/block.hpp
```diff
@@ -157,19 +158,20 @@ class BC_API block
     bool is_extra_coinbases() const;
     bool is_final(size_t height, uint32_t block_time) const;
     bool is_distinct_transaction_set() const;
-    bool is_valid_coinbase_claim(size_t height) const;
+    bool is_valid_coinbase_claim(size_t height, const settings& settings) const;
```
evoskuil: Let's pass the two necessary settings value to `subsidy`, `reward` and `is_valid_coinbase_claim`. This will result in one more parameter for the latter two methods, but I'd like to stick to the approach of not using settings for methods/functions. However we can make an exception for `check`, `accept`, and `connect`.

include/bitcoin/bitcoin/chain/block.hpp
```diff
@@ -157,19 +158,20 @@ class BC_API block
...
     bool is_forward_reference() const;
     bool is_internal_double_spend() const;
     bool is_valid_merkle_root() const;
     bool is_segregated() const;
 
-    code check(bool retarget) const;
-    code check_transactions() const;
-    code accept(bool transactions=true, bool header=true) const;
-    code accept(const chain_state& state, bool transactions=true,
+    code check(bool retarget, const settings& settings) const;
```
evoskuil: For consistency (see tx) the `retarget` parameter should be second and default to `true`

test/settings.cpp
```diff
@@ -46,6 +46,22 @@ BOOST_AUTO_TEST_CASE(settings__construct__default_context__expected)
     BOOST_REQUIRE_EQUAL(configuration.bip9_version_bit0, 1u);
     BOOST_REQUIRE_EQUAL(configuration.bip9_version_bit1, 2u);
     BOOST_REQUIRE_EQUAL(configuration.bip9_version_base, 0x20000000);
+    BOOST_REQUIRE_EQUAL(configuration.satoshi_per_bitcoin, 100000000);
+    BOOST_REQUIRE_EQUAL(configuration.initial_block_subsidy_bitcoin, 50);
+    BOOST_REQUIRE_EQUAL(configuration.recursive_money, 9999999989u);
+    BOOST_REQUIRE_EQUAL(configuration.retarget_subsidy_interval, 210000);
+    BOOST_REQUIRE_EQUAL(configuration.no_retarget_subsidy_interval, 150);
+    const auto recursive_money = configuration.max_money_recursive(
+        configuration.initial_block_subsidy_satoshi());
```
evoskuil: Don't wrap test code.

include/bitcoin/bitcoin/settings.hpp
toxeus: I feel the same about logic inside the settings class. That's why I explicitly asked you on slack about where these functions should go before starting to work on this PR. You've told me that it's ok for now to move them into the settings class :man_shrugging:

include/bitcoin/bitcoin/chain/block.hpp
toxeus: Will do :+1:

include/bitcoin/bitcoin/chain/block.hpp
toxeus: makes sense

include/bitcoin/bitcoin/settings.hpp
toxeus: :+1:

test/settings.cpp
toxeus: Oh, I thought that it's only code inside the `BOOST_*` macros that shouldn't be wrapped.

include/bitcoin/bitcoin/settings.hpp
evoskuil: I suppose now is no longer then :?

test/settings.cpp
evoskuil: https://github.com/libbitcoin/libbitcoin/wiki/Test-Guide#line-length

evoskuil: Exception exists for declaring complex structures.

toxeus: For my benefit, what's the rationale for this rule?

evoskuil: If you read the whole test guide there is an emphasis on declarativeness. For the same reason no conditions are allowed in the tests. Since tests are not themselves tested (infinite regression problem), it is critical that they can be evaluated for correctness visually by humans. For this reason we are strict about reducing visual clutter.

toxeus: thx :+1:

src/settings.cpp
```diff
@@ -233,4 +239,21 @@ settings::settings(config::settings context)
     }
 }
 
+uint64_t settings::bitcoin_to_satoshi(uint64_t bitcoin_uints) const
```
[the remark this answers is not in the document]
toxeus: I don't know, [you](https://github.com/libbitcoin/libbitcoin/commit/24d4d9e3f9aa57f58693e8390568a555f7c9ad73#diff-995071d381795f41bb69db16e0138173L113) tell me ;)

src/settings.cpp
```diff
@@ -233,4 +239,21 @@ settings::settings(config::settings context)
...
+// CONSENSUS: This is the true maximum amount of money that can be created.
+// The satoshi client uses a "sanity check" value that is effectively based on
+// a round but incorrect value of recursive_money, which is higher than this
+// true value. Despite comments to the contrary in the satoshi code, no value
+// could be consensus critical unless it was *less* than the true value.
+//**************************************************************************
+uint64_t settings::max_money() const
+{
+    return recursive_money * subsidy_interval;
+}
```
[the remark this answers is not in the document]
toxeus: Good catch! All the noise from the refactorings made me stop seeing the elegant simplifications. Thx!

src/settings.cpp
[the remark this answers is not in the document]
evoskuil: I was, it's a typo.

toxeus:
Updated PR and propagated changes into all remaining repos.

Issue #1006, opened 2018-08-11
Revise readme to recommend prefix, static, and build dependencies.


evoskuil:

evoskuil:
Discussion: https://github.com/libbitcoin/libbitcoin/issues/1005

PR #1009, opened 2018-08-14
Update readme to include recommended options (such as prefix/static)

thecodefactory:
Addresses #1006 

Might be a bit heavy-handed, but recommends always using a prefix, building boost (instead of recommending a system install) and using static linkage over shared when using install.sh

This still points to version3 installer since version4 is not yet production.

Once we hammer out what we want for the options, I can update other repos (including v3).

### DIFF
--- a/README.md
+++ b/README.md
@@ -61,21 +61,17 @@ Next install the [build system](http://wikipedia.org/wiki/GNU_build_system) (Aut
 ```sh
 $ sudo apt-get install build-essential autoconf automake libtool pkg-config git
 ```
-Next install the [Boost](http://www.boost.org) (minimum 1.57.0) development package:
-```sh
-$ sudo apt-get install libboost-all-dev
-```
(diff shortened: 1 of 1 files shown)


evoskuil:
The reason I used `/home/me/myprefix` was that it implies a directory that doesn't require sudo.

evoskuil:
You should not use `--enable-static` (or `--enable_shared`) with the installer, as both are enabled by default. The [relevant install.sh options](https://github.com/libbitcoin/libbitcoin/blob/master/install.sh#L40-L41) are:
```
# --disable-shared         Disables shared library builds.
# --disable-static         Disables static library builds.
```
The installer converts these to the appropriate parameterizations.

thecodefactory:
> You should not use --enable-static (or --enable_shared) with the installer, as both are enabled by default

Habit I guess ;-)  I've removed those, but left in the --disable-shared option.

evoskuil:
Yeah, it’s necessary in configure in some context, but we make that simpler for install.

evoskuil:
Please make matching changes in bs/bx repos.

PR #1013, opened 2018-08-15
Make config::block default constructible.

toxeus:
`config::block` needs to be default constructible such that it can be used with `boost::program_options`.


toxeus:
Please, don't merge. I need to propagate this change through all repos.

toxeus:
@evoskuil alright, I have opened PRs in all dependent repos. It would be good to merge this before it conflicts with other changes.

PR #1014, opened 2018-08-17
Add copy assignment operator for config::block.

toxeus:
This is the last bit needed to make `config::block` work using `boost::program_options`. It's necessary to implement the copy assignment explicitly because the copy assignment operator for `chain::block`, which is its internal type, has been deleted.

### DIFF
--- a/src/config/block.cpp
+++ b/src/config/block.cpp
@@ -49,6 +49,12 @@ block::block(const block& other)
 {
 }
 
+block& block::operator=(const block& other)
+{
+    value_ = chain::block(other.value_);
+    return *this;
+}
+
 block& block::operator=(chain::block&& other)
 {
     value_ = std::move(other);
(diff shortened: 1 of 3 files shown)


src/config/block.cpp
```diff
@@ -49,6 +49,13 @@ block::block(const block& other)
 {
 }
 
+block& block::operator=(const block& other)
+{
+    chain::block value = block(other);
+    value_ = std::move(value);
```
evoskuil:
```cpp
    value_ = block(other);
```

evoskuil: I assume the move is required to be explicit here, as opposed to simply relying on the r-value?

toxeus: Exactly, I thought that like this it's more descriptive and it clearly shows that I'm forced to do a move assignment here. But your comment proves that this rather adds confusion than clarity. I'll change it such that it relies on the r-value.

evoskuil:
It’s just that move can sometimes be expensive too.

toxeus:
True, but for config classes performance is not that critical.

evoskuil:
Yes, but avoid when not necessary.

PR #1015, opened 2018-08-21
Add litecoin hard fork.

toxeus:
The litecoin hard fork includes:

* scrypt as PoW hashing algorithm
* fix for Satoshi's off-by-one bug for timespan computation
* workaround for uint256_t overflow when new target is computed

The implementation has been kept consistent with the regtest hard fork implementation where a `retarget` bool is passed around.

I have successfully tested this using `bn`.


src/chain/header.cpp
```diff
@@ -430,6 +430,8 @@ hash_digest header::hash() const
     mutex_.unlock_upgrade();
     ///////////////////////////////////////////////////////////////////////////
 
+    if (litecoin)
+        return litecoin_hash(to_data());
```
toxeus: For now, I've omitted here caching litecoin's pow hash because I wan't sure if that benefits performance.

src/chain/chain_state.cpp
```diff
@@ -513,7 +524,13 @@ chain_state::data chain_state::to_pool(const chain_state& top,
     // If promoting from retarget height, move that timestamp into retarget.
     if (retarget &&
         is_retarget_height(height - 1u, settings.retargeting_interval))
-        data.timestamp.retarget = data.timestamp.self;
+    {
+        // litecoin fixed Satoshi's off-by-one retargeting bug.
+        if (!litecoin || height == 1)
+            data.timestamp.retarget = data.timestamp.self;
+        else
+            data.timestamp.retarget = *next(data.timestamp.ordered.crbegin());
```
toxeus: I should change this to `std::next`.

src/chain/chain_state.cpp
```diff
@@ -317,8 +321,14 @@ uint32_t chain_state::work_required_retarget(const data& values,
     BITCOIN_ASSERT_MSG(!bits.is_overflowed(), "previous block has bad bits");
 
     uint256_t target(bits);
+    const bool needs_shift = msb(target) > msb(pow_limit) - 1;
```
toxeus: I should use `auto` here.

include/bitcoin/bitcoin/machine/rule_fork.hpp
```diff
@@ -70,6 +70,9 @@ enum rule_fork : uint32_t
     /// Prevent dummy value malleability (soft fork, feature).
     bip147_rule = 1u << 13,
 
+    /// Use litecoin blockchain validation (hard fork, litecoin).
+    litecoin = 1u << 14,
```
evoskuil: Don't refer to specific coins, indicate each independent behavior change with an independent flag, e.g.:
```ini
[fork]
retarget_overflow_patch = false
script_proof_of_work = false
time_warp_patch =  false
```
and pass through code in this manner.

src/chain/block.cpp
```diff
@@ -806,14 +806,16 @@ code block::connect_transactions(const chain_state& state) const
...
+    uint32_t no_retarget_proof_of_work_limit, bool retarget, bool litecoin)
+    const
 {
     metadata.start_check = asio::steady_clock::now();
 
     code ec;
 
     if ((ec = header_.check(timestamp_future_seconds,
-        retarget_proof_of_work_limit, no_retarget_proof_of_work_limit, retarget)
+        retarget_proof_of_work_limit, no_retarget_proof_of_work_limit, retarget,
```
evoskuil: I don't understand why both retarget_proof_of_work_limit and no_retarget_proof_of_work_limit parameters still exist. This should have been eliminated by moving the limit to config.

src/chain/chain_state.cpp
evoskuil: use ternary:
```cpp
        data.timestamp.retarget = (time_warp_patch && height != 0) ? *next(data.timestamp.ordered.crbegin()) : 
            data.timestamp.self;
```

include/bitcoin/bitcoin/chain/header.hpp
```diff
@@ -151,7 +151,7 @@ class BC_API header
     uint32_t nonce() const;
     void set_nonce(uint32_t value);
 
-    hash_digest hash() const;
+    hash_digest hash(bool litecoin=false) const;
```
evoskuil: This is getting ugly.

src/chain/block.cpp
toxeus: Agreed. It slipped through in the massive refactoring PRs. Will fix it.

src/chain/chain_state.cpp
evoskuil: Incorporate flag into this result:

```cpp
const auto patch = script::is_enabled(forks, rule_fork::retarget_overflow_patch);
const auto shift = patch && (msb(target) > msb(pow_limit) - 1);
```

This is simpler code, prevents one condition and four math operations if the flag is not enabled.

is this safe from underflow?
```
msb(pow_limit) - 1
```

What namespace and header is `msb`?

include/bitcoin/bitcoin/chain/header.hpp
toxeus: As I wrote in the commit msg, I was trying to stay consistent with how it's done for the `retarget` parameter where [similar ugliness](https://github.com/libbitcoin/libbitcoin/blob/81b61747ec29a45ba43f08c437c5fa06d3152ebd/include/bitcoin/bitcoin/chain/block.hpp#L167) exists. Do you prefer a separate function then?

src/chain/chain_state.cpp
```diff
@@ -317,8 +321,14 @@ uint32_t chain_state::work_required_retarget(const data& values,
...
+    const auto litecoin = script::is_enabled(forks, rule_fork::litecoin);
+    if (litecoin && needs_shift)
+        target >>= 1;
```
evoskuil:
```cpp
target >>= (shift ? 1 : 0)
```

src/chain/chain_state.cpp
```diff
@@ -317,8 +321,14 @@ uint32_t chain_state::work_required_retarget(const data& values,
...
     target *= retarget_timespan(values, min_timespan, max_timespan);
     target /= target_timespan_seconds;
+    if (litecoin && needs_shift)
+        target <<= 1;
```
evoskuil: target <<= shift;

src/chain/header.cpp
evoskuil: The benefits of caching the hash are significant. However it's not clear here whether the scrypt hash is used for anything other than PoW. If it is not then this method should not be modified. If it is then there will be a lot more impact.

toxeus: It's used for PoW exclusively.

evoskuil:
Have you reviewed cancoin's litecoin fork?

toxeus:
Yes, I did.

include/bitcoin/bitcoin/chain/header.hpp
evoskuil: Parameterizing hash() creates a lot of potential issues.

src/chain/header.cpp
evoskuil: Then don't parameterize hash, just condition the PoW calc. Otherwise it will break the other stuff that hash is used for.

include/bitcoin/bitcoin/machine/rule_fork.hpp
evoskuil: We may of course encounter other scrypt pow algorithms, and other patches to these issues. There is a discussion in Bitcoin no regarding the time warp fix. I'd prefer to cross that bridge when we come to it rather than integrating coin names.

toxeus: I fully agree that it makes sense to avoid using coin names :+1:

src/chain/chain_state.cpp
toxeus: https://github.com/libbitcoin/libbitcoin/pull/1018

src/chain/block.cpp
toxeus: https://github.com/libbitcoin/libbitcoin/pull/1016

src/chain/chain_state.cpp
toxeus: > is this safe from underflow?

No, it's not. But that's a big edge case since it'll only underflow when the user chooses `bc::settings.proof_of_work_limit` to represent `uint256_t(1)` (e.g. 0x1010000). I've never seen a block that met this difficulty requirement :wink:. However, it's trivial to fix the underflow so I'll do it.

evoskuil: Isn’t it a consensus question?

toxeus: This is not easy to answer: if we take ltc or ftc then this will never underflow for what they've chosen as `bc::settings.proof_of_work_limit`. Moreover, what you can see here is exactly what they've implemented. But nowadays we have so many coins - I'm not able to answer this for all of them :man_shrugging:

evoskuil: Okay, makes sense. Pretty obvious, just didn't look too closely. I would use this then:
```cpp
const auto shift = patch && (msb(target) > safe_subtract(msb(pow_limit), 1) ? 1u : 0u;
```
This will throw if it underflows. FWIW we try to never allow potential under/overflow without explicit handling (or a comment as to why it is correct behavior). So any time I see math I question it.

evoskuil: This is problematic now that it is parameterized:
```cpp
static const uint256_t pow_limit(compact{ retarget_proof_of_work_limit });
```
The `retarget_proof_of_work_limit` parameter can change but the result will not.

PR #1023, opened 2018-08-27
Encapsulate currency unit settings.

toxeus:


src/settings.cpp
```diff
@@ -41,7 +41,8 @@ settings::settings()
     bip9_version_base(0x20000000),
     satoshi_per_bitcoin(100000000),
     initial_block_subsidy_bitcoin(50),
-    recursive_money(9999999989u)
+    recursive_money(max_money_recursive(bitcoin_to_satoshi(
+        initial_block_subsidy_bitcoin)))
```
evoskuil: This can’t be safely exposed as a property unless computed (vs. initialized). Same holds for other compued initializatuons above.

toxeus:
I was waiting for the PR with the renaming to be merged before continuing here because both manipulate the settings.cpp file and hence might conflict. The renaming has a higher prio because it blocked the parser PR in the node repo. I have updated the PR now.

toxeus:
As discussed on slack, I have properly encapsulated the currency unit settings. I have successfully tested the setter methods with boost::program_options and can confirm that it works as expected. Other bc::settings members that are derived will be cleaned up in upcoming PRs.

src/settings.cpp
```diff
@@ -237,6 +236,25 @@ settings::settings(config::settings context)
         {
         }
     }
+    max_money_ = recursive_money_ * subsidy_interval_;
+}
+
+uint64_t settings::bitcoin_to_satoshi(uint64_t bitcoin_units) const
+{
+
```
evoskuil: Style: remove whitespace.

src/settings.cpp
```diff
@@ -237,6 +236,25 @@ settings::settings(config::settings context)
         {
         }
     }
+    max_money_ = recursive_money_ * subsidy_interval_;
+}
+
+uint64_t settings::bitcoin_to_satoshi(uint64_t bitcoin_units) const
```
evoskuil: For property methods I generally just use "value" for the parameter since the parameter name is redundant with the property name.

include/bitcoin/bitcoin/settings.hpp
```diff
@@ -40,6 +40,11 @@ class BC_API settings
     //--------------------------------------------------------------------------
 
     uint64_t bitcoin_to_satoshi(uint64_t bitcoin_units=1) const;
+    void initial_block_subsidy_bitcoin(uint64_t initial_block_subsidy_bitcoin);
+    uint64_t initial_block_subsidy_bitcoin() const;
+    void subsidy_interval(uint64_t subsidy_interval);
+    uint64_t subsidy_interval() const;
+    uint64_t max_money() const;
```
evoskuil: It seems inevitable when creating a class with fields that eventually proper behavior will drive it into properties.

src/settings.cpp
toxeus: Done.

evoskuil:
Looks good, please merge when ready.

PR #1031, opened 2018-09-09
Installer readability improvement

ldz1:
Hello there,

I believe the readability of the install.sh can be improved. It would be easier to read the script from top to bottom instead of jumping between different sections. The other small improvements are mentioned in the commit description.

Best regards,
ldz1

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -53,433 +53,255 @@
 
 # Define constants.
 #==============================================================================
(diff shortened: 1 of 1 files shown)


evoskuil:
This is generated code and as such any change must be applied at libbitcoin-build.

PR #1032, opened 2018-09-09
A simple shutdown_ flag to avoid shutdown deadlock

JasonCoombs:
This may be only for discussion. It may be better for programmers to know that they can't ever allow threads in the threadpool to ever be re-used for new calls to spawn new work until after any call to join() first completes successfully, but we can easily defend against a deadlock during shutdown in this manner and probably not leave database tables in a corrupted state. Even when we're 100% sure that all possible deadlock conditions have been avoided in our code, we may want this shutdown_ flag to guard against deadlocks when other developers build their own solutions using libbitcoin libraries and inadvertently allow new work to spawn on a threadpool during a join() operation. Because so many instances of shutdown deadlocks are going to occur as a result of unpredictable network events and buffers that are still being drained and dispatched to handler functions, it seems like a good idea to have this shutdown flag and to allow network buffers that might otherwise produce perfectly useful state changes in the database (if not for imperfect handling of the join() condition in all pending work queues) to be discarded instead of being added to the database because of comprehensive queue draining just prior to shutdown.

Also, while libbitcoin-node and libbitcoin-server in the original use case (the Bitcoin protocol and Core blockchain network) may be, by design, meant to launch and continue running for long periods of time, in new applications libbitcoin libraries may be put to work within servers and cluster nodes that are constantly launching and shutting down. There is room for improvement in the way that shutdown is handled, generally, to reduce the number of database corruption events especially during the evolution of new cryptoeconomics applications and communities.

### DIFF
--- a/src/utility/threadpool.cpp
+++ b/src/utility/threadpool.cpp
@@ -28,6 +28,7 @@ namespace libbitcoin {
 threadpool::threadpool(size_t number_threads, thread_priority priority)
   : size_(0)
 {
+    shutdown_ = false; // prevents deadlock due to threads_ that may spawn during shutdown
     spawn(number_threads, priority);
 }
 
@@ -52,45 +53,51 @@ size_t threadpool::size() const
 // This is not thread safe.
 void threadpool::spawn(size_t number_threads, thread_priority priority)
 {
-    // This allows the pool to be restarted.
-    service_.reset();
+    if(!shutdown_)
+    {
+        // This allows the pool to be restarted.
+        service_.reset();
 
-    for (size_t i = 0; i < number_threads; ++i)
-        spawn_once(priority);
+        for (size_t i = 0; i < number_threads; ++i)
+            spawn_once(priority);
+    }
 }
 
 void threadpool::spawn_once(thread_priority priority)
 {
-    ///////////////////////////////////////////////////////////////////////////
-    // Critical Section
-    work_mutex_.lock_upgrade();
-
-    // Work prevents the service from running out of work and terminating.
-    if (!work_)
+    if(!shutdown_)
     {
-        work_mutex_.unlock_upgrade_and_lock();
-        //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
-        work_ = std::make_shared<asio::service::work>(service_);
+        ///////////////////////////////////////////////////////////////////////////
+        // Critical Section
+        work_mutex_.lock_upgrade();
 
-        work_mutex_.unlock_and_lock_upgrade();
-        //-----------------------------------------------------------------
+        // Work prevents the service from running out of work and terminating.
+        if (!work_)
+        {
+            work_mutex_.unlock_upgrade_and_lock();
+            //+++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++++
+            work_ = std::make_shared<asio::service::work>(service_);
+
+            work_mutex_.unlock_and_lock_upgrade();
+            //-----------------------------------------------------------------
+        }
+
+        work_mutex_.unlock_upgrade();
+        ///////////////////////////////////////////////////////////////////////////
+
+        ///////////////////////////////////////////////////////////////////////////
+        // Critical Section
+        unique_lock lock(threads_mutex_);
+
+        threads_.push_back(asio::thread([this, priority]()
+        {
+            set_priority(priority);
+            service_.run();
+        }));
+
+        ++size_;
+        ///////////////////////////////////////////////////////////////////////////
     }
-
-    work_mutex_.unlock_upgrade();
-    ///////////////////////////////////////////////////////////////////////////
-
-    ///////////////////////////////////////////////////////////////////////////
-    // Critical Section
-    unique_lock lock(threads_mutex_);
-
-    threads_.push_back(asio::thread([this, priority]()
-    {
-        set_priority(priority);
-        service_.run();
-    }));
-
-    ++size_;
-    ///////////////////////////////////////////////////////////////////////////
 }
 
 void threadpool::abort()
@@ -102,6 +109,7 @@ void threadpool::shutdown()
 {
     ///////////////////////////////////////////////////////////////////////////
     // Critical Section
+    shutdown_ = true; // prevents deadlock due to threads_ that may spawn during shutdown
     unique_lock lock(work_mutex_);
 
     work_.reset();


src/utility/threadpool.cpp
```diff
@@ -52,45 +53,51 @@ size_t threadpool::size() const
...
 
 void threadpool::spawn_once(thread_priority priority)
 {
-    ///////////////////////////////////////////////////////////////////////////
-    // Critical Section
-    work_mutex_.lock_upgrade();
-
-    // Work prevents the service from running out of work and terminating.
-    if (!work_)
+    if(!shutdown_)
```
evoskuil: Style: `if` is not a function:
```cpp
if (!shutdown_)
```

src/utility/threadpool.cpp
```diff
@@ -28,6 +28,7 @@ namespace libbitcoin {
 threadpool::threadpool(size_t number_threads, thread_priority priority)
   : size_(0)
 {
+    shutdown_ = false; // prevents deadlock due to threads_ that may spawn during shutdown
```
evoskuil: Style: no trailing comments. Move to initializer list.

src/utility/threadpool.cpp
```diff
@@ -52,45 +53,51 @@ size_t threadpool::size() const
 // This is not thread safe.
 void threadpool::spawn(size_t number_threads, thread_priority priority)
 {
-    // This allows the pool to be restarted.
-    service_.reset();
+    if(!shutdown_)
```
evoskuil: Style: if is not a function.

src/utility/threadpool.cpp
```diff
@@ -102,6 +109,7 @@ void threadpool::shutdown()
 {
     ///////////////////////////////////////////////////////////////////////////
     // Critical Section
+    shutdown_ = true; // prevents deadlock due to threads_ that may spawn during shutdown
```
evoskuil: Either move into actual critical section or move above critical section comment.

evoskuil: Thread safety: this write is not atomic with the reads, making the implementation not thread safe.

evoskuil:
> constantly launching and shutting down

The PR doesn't address this scenario, and it wouldn't matter how many times this was happening, an issue would exist even on one shutdown. The problem is launching threads after join. This is not a supported use of the threadpool. Guarding against it is fine, but it has nothing to do with the frequency of startup and shutdown. It simply requires that shutdown not happen during startup.

JasonCoombs:
Very helpful review, thank you. Especially appreciate the style guidance, I will try harder to follow this
https://github.com/libbitcoin/libbitcoin/wiki/Style-Guide

Note that the example of keyword vs function call, here, should appear on the next line for wiki clarity:

`Keywords such as if and for are not functions, one space before left parenthesis: while (!done) ....`

I see where spawn() is called only once, on startup, in p2p.cpp here:
https://github.com/libbitcoin/libbitcoin-network/blob/master/src/p2p.cpp#L101

Do we not have any mechanism for increasing the number of threads after startup? The config setting description suggests that the number of threads is a minimum, or a target number, not a maximum and a fixed number of threads. Should that terminology be revised to more clearly indicate that the config settings dictate fixed numbers of threads?

If a shutdown_ flag is added as proposed here, I don't think there is a deadlock risk or concurrency issue on reads that receive a 'false' after the shutdown function has been called but before shutdown_ is set to 'true' unless it is possible for an existing thread to call spawn() or spawn_once() and fail to reach the point where a unique lock is obtained on threads_mutex_ before the thread that called shutdown() finishes that call and makes the subsequent call to join() and locks the mutex first.

Where a deadlock may emerge (only in the future when other developers are adapting libbitcoin to custom applications and do have new threads being added to a threadpool dynamically) is here:

https://github.com/libbitcoin/libbitcoin/blob/master/src/utility/threadpool.cpp#L115

attempting to spawn a new thread (adding to the threadpool) after a thread has called join() would cause a deadlock (not the ones that are being debugged presently as a result of v4 refactoring) here:

https://github.com/libbitcoin/libbitcoin/blob/master/src/utility/threadpool.cpp#L84

... if the new thread is spawned by one of the existing threads being joined. This is what I thought was happening in the shutdown deadlock that I was debugging, because it appeared there were two threads waiting on the same unique_lock but you're right that a new thread being spawned during shutdown isn't the source of the second unique_lock, that must be pending on an internal BOOST mutex not waiting on threads_mutex_ in spawn_once().

Couldn't we lock the threads_mutex_ before locking the work_mutex_ in spawn_once() and thereby eliminate any possible race condition with respect to future dynamic thread spawn calls which read a 'false' from shutdown_ and then are preempted while another thread (on the same or a different core) that has called shutdown() also ends up reaching its call to join() and locking threads_mutex_ before the spawn_once() thread gets allocated more CPU time so that it can lock threads_mutex_ and add a new thread to the threadpool before the join() call needs the lock?

JasonCoombs:
According to Boost.Asio documentation:

"Note that all threads that have joined an io_context's pool are considered equivalent, and the io_context may distribute work across them in an arbitrary fashion."

"Asynchronous completion handlers will only be called from threads that are currently calling io_context::run()."

https://www.boost.org/doc/libs/1_68_0/doc/html/boost_asio/overview/core/threads.html

"* This function will block until the thread has exited."

https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/thread.hpp#L71

For as long as a thread is still waiting on the boost::asio::io_context::run() function to return, it may be reused arbitrarily as a member of the threadpool for invoking handler functions. But it seems likely that no thread that has already been join()'ed in our join() function could be reused by the io_context, as it will have exited.

Note: io_context is the new name for io_service

https://www.boost.org/doc/libs/1_68_0/doc/html/boost_asio/reference/io_service.html

"Typedef for backwards compatibility."

`typedef io_context io_service;`

JasonCoombs:
This, also, for future consideration if a more forceful shutdown is helpful which discards pending handler function calls:

https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/io_context.hpp#L164

and

https://github.com/chriskohlhoff/asio/blob/master/asio/include/asio/io_context.hpp#L463

`/**
   * This function does not block, but instead simply signals the io_context to
   * stop. All invocations of its run() or run_one() member functions should
   * return as soon as possible. Subsequent calls to run(), run_one(), poll()
   * or poll_one() will return immediately until restart() is called.
   */`

JasonCoombs:
I've added a shutdown mutex in my development branch and will re-submit the two PRs as one PR with that shutdown mutex also. On the subject of discarding handlers, I'm still trying to fully understand what you mean by "we cannot discard handlers" -- what I meant to say was that we do appear to be able to cancel the loop that keeps spawning new handlers until the entire pending candidate chain is processed for potential validation. Shutting down in the middle of that loop doesn't seem to cause any problem, as valid state transitions have completed and the loop which spawns new handlers resumes on next startup.

evoskuil:
Please reopen as a single (i.e. non-breaking) PR.

PR #1033, opened 2018-09-09
A simple shutdown_ flag to avoid shutdown deadlock

JasonCoombs:
This adds the shutdown_ bool flag as private.


evoskuil:
Please reopen as a single (i.e. non-breaking) PR.

PR #1050, opened 2018-10-27
witness param supported in to_data(witness)

JasonCoombs:
this was a bug in v3 also, if anyone had ever tried to retrieve witness data with block.to_data(true)


evoskuil:
Witness parameterization is lacking in unit test coverage.

PR #1051, opened 2018-10-27
witness param supported in to_data(witness)

JasonCoombs:
if anyone had ever tried to retrieve witness data with block.to_data(true) it wouldn't have included witness data. now it does.


JasonCoombs:
see also https://github.com/libbitcoin/libbitcoin/pull/1050

evoskuil:
Witness parameterization is lacking in unit test coverage.

PR #1063, opened 2018-11-07
Verifier of backend manager was improved.

ldz1:
Previous implementation doesn't detect missing ICU. boost:normalize fails silently without ICU.


evoskuil:
See Windows test failures. If these are correct we should probably resolve the dependency issue before merging.

ldz1:
Yes, I expected it :). I agree.

thecodefactory:
For any of the mnemonic work, this appears to be pretty high priority.  Any ideas on resolution?

evoskuil:
Disabled WITH_ICU in VS builds :/.

Issue #1060, opened 2018-11-05
Dictionary (and tests) possibly not encoded consistent with reference.


evoskuil:
Reference [0x61cc81]:
https://github.com/bitcoin/bips/blob/master/bip-0039/spanish.txt#L93

Dictionary [0xc3a1]:
https://github.com/libbitcoin/libbitcoin/blob/master/src/wallet/dictionary.cpp#L2174

evoskuil:
https://github.com/libbitcoin/libbitcoin/pull/1059

thecodefactory:
See https://github.com/libbitcoin/libbitcoin/pull/1073

I can't imagine it was intentional, but the BIP-39 reference word list encoding is *not* identical to what Electrum uses for both Spanish and Japanese.  Since they state that they share the same word list, we previously used the same word lists for both implementations, which simply doesn't work.

evoskuil:
Then we should split out all languages, even en, to avoid confusion and assumptions.

thecodefactory:
@evoskuil Update on this -- part of this work discovered that we had a bug in BIP39 that required pre-normalized input mnemonics, rather than normalizing them internally (though passphrase normalization was properly performed).  The electrum dictionaries that differ are in encoding only, meaning that the words are the same and that after normalization the BIP39 wordlist and electrum wordlist match.  I've added test cases to confirm this locally since they are currently split out.

What this means is that we should be able to collapse the dictionaries back to one copy (using the non-normalized, electrum version), while ensuring that BIP39 internally normalizes (electrum always required it, so it wasn't an issue there).

It also means the Windows normalization bug was present all along (where ICU failed), but it didn't present on BIP39 testing because the input word lists were already in the proper normalized form.  Externally entered user input could have always failed to find the entered words, meaning validation could have failed on valid BIP39 mnemonics due purely to the encoding issues.  This still doesn't affect english mnemonics, where ICU isn't required. 


The below test won't be needed when I collapse back the dictionaries, but it's presented to demonstrate:

```
BOOST_AUTO_TEST_CASE(electrum__test_dictionary_normalization_differences__valid)
{   
    const auto dictionary_en = language::en;
    const auto electrum_dictionary_en = language::electrum::en;

    const auto dictionary_ja = language::en;
    const auto electrum_dictionary_ja = language::electrum::en;

    BOOST_REQUIRE(dictionary_en.size() == dictionary_size);
[...]

PR #1067, opened 2018-11-13
Update src/wallet/dictionary.cpp

ebCrypto:
with regards to #1060 
I copied the source file https://raw.githubusercontent.com/bitcoin/bips/master/bip-0039/spanish.txt adding the necessary "


thecodefactory:
Re-ordering the english can't be correct.  I'm not sure the dictionaries need adjustment either.  Any further info provided there?

thecodefactory:
> I grabbed the Spanish words with special characters from the bip-0039 source file

The original word lists were taken from:

https://github.com/bitcoin/bips/blob/master/bip-0039/bip-0039-wordlists.md.

Can you clarify what the difference is?

thecodefactory:
@ebCrypto I'm going to close this in favor of #1073 

Let me know if you think it doesn't address the issue and we can revisit.

ebCrypto:
Ok it makes sense!

PR #1073, opened 2018-11-14
Electrum normalization fix

thecodefactory:


thecodefactory:
This depends on proper Windows ICU support, possibly related to #1063 

Alternatively, it could just not be supported on Windows and unavailable there, which is the inferior option.

PR #1080, opened 2018-12-08
ec_scalar: add non-const accessor and construct by value

narodnik:
As the title says.

non-const accessor is needed to be able to modify underlying secret value without unecessary copying.

construct by value is very useful for doing calculations where we blind output amounts (for example).

Also the unit test filenames for ec_point and ec_scalar were reversed. I switched them over.

### DIFF
--- a/src/math/ec_scalar.cpp
+++ b/src/math/ec_scalar.cpp
@@ -38,6 +38,14 @@ ec_scalar::ec_scalar(const ec_secret& secret)
 {
 }
 
+ec_scalar::ec_scalar(uint64_t value)
+  : valid_(true)
+{
+    secret_ = zero;
+    auto serial = make_unsafe_serializer(secret_.end() - 4);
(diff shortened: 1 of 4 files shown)


src/math/ec_scalar.cpp
```diff
@@ -38,6 +38,14 @@ ec_scalar::ec_scalar(const ec_secret& secret)
 {
 }
 
+ec_scalar::ec_scalar(uint64_t value)
+  : valid_(true)
+{
+    secret_ = zero;
+    auto serial = make_unsafe_serializer(secret_.end() - 4);
+    serial.write_4_bytes_big_endian(value);
```
evoskuil: Writing 4 bytes casted from an 8 byte value, as big endian, into the high order bytes of a 32 byte value appears completely arbitrary and unrelated to the behavior of an ec scalar.

PR #1090, opened 2018-12-29
Remove dead code and re-enable some uint256 tests.

thecodefactory:


evoskuil:
When I switched to using boost for `uint256_t` I was a little uncomfortable getting rid of these tests. It doesn't really make sense to test boost, but the expected behavior is consensus critical and not necessarily obvious. Maybe enable them?

thecodefactory:
> When I switched to using boost for `uint256_t` I was a little uncomfortable getting rid of these tests. It doesn't really make sense to test boost, but the expected behavior is consensus critical and not necessarily obvious. Maybe enable them?

Ok, was not sure of the intent here.  I can close this, given this.

evoskuil:
We should get rid of the dead code tho, maybe just enable the tests.

PR #1103, opened 2019-01-25
Add a method to return the output script from a payment address.

thecodefactory:
@evoskuil I think something like this will help simplify generating the output index hashes.  It's also directly applicable to a command like `script-from-address` (in bx).  Thoughts?


evoskuil:
Sure

PR #1114, opened 2019-05-20
Typo in src

ldz1:


ldz1:
I have to remove all commits except the last.

evoskuil:
Thanks for this PR, but you should not make changes here to generated files. Those must be changed via the templates in libbitcoin-build, which will cause them to be updated in the next source generation.

ldz1:
@evoskuil which generated files do you mean? "Files changed" contain almost exclusively the files from src.

Besides, I have merged instead of rebase and I suppose you would like to have one smooth commit. I can try to change this soon if desired.

evoskuil:
See file comments at the top of the files for what is generated.

Clean commits would be better. Let us know if you need any more help with that.

ldz1:
@evoskuil: As I mentioned before on slack, the commits didn't affect the generated files in the result. 

Nevertheless, I have adapted it and I hope now it is clearer. (Thanks for the cherry-pick hint)

evoskuil:
Yes, I see that now. Must have looked at a merged file last time. This PR is much better. I travel a lot and work from my phone, so easy to get confused with big commits.

PR #1115, opened 2019-06-03
Remove unused include from src/chain and src/config

ldz1:


thecodefactory:
@ldz1 Thanks, looks good!

PR #1117, opened 2019-06-04
Multiple small changes in block (redudant code, const etc.)

ldz1:
Remove: Redundant member initializer in constructor initialization list.
https://clang.llvm.org/extra/clang-tidy/checks/readability-redundant-member-init.html

Remove: std::move has no effect (trivially copyable)
https://clang.llvm.org/extra/clang-tidy/checks/performance-move-const-arg.html

Remove: Redundant qualifier

Remove: Redundant else

Remove: Unused include

Add: const

Change: Hide warning "using the result of an assignment as a condition" through parenthesis.

### DIFF
--- a/src/chain/block.cpp
+++ b/src/chain/block.cpp
@@ -60,9 +60,7 @@ using namespace boost::adaptors;
 //-----------------------------------------------------------------------------
 
 block::block()
-  : metadata{},
-    header_(),
-    transactions_()
+  : metadata{}
 {
 }
 
@@ -388,7 +386,7 @@ size_t block::locator_size(size_t top)
 block::indexes block::locator_heights(size_t top)
 {
     size_t step = 1;
-    block::indexes heights;
+    indexes heights;
     const auto reservation = locator_size(top);
     heights.reserve(reservation);
 
@@ -449,8 +447,8 @@ uint64_t block::subsidy(size_t height, uint64_t subsidy_interval,
 size_t block::signature_operations() const
 {
     const auto state = header_.metadata.state;
-    const auto bip16 = state->is_enabled(rule_fork::bip16_rule);
-    const auto bip141 = state->is_enabled(rule_fork::bip141_rule);
+    const auto bip16 = state->is_enabled(bip16_rule);
+    const auto bip141 = state->is_enabled(bip141_rule);
     return state ? signature_operations(bip16, bip141) : max_size_t;
 }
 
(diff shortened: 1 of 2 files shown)


ldz1:
I have changed my gnu c++ compiler from 7 to 4.8.5 and I am able to build. The 4.8.4 c++ compiler on the build server seems to be ugly :D. I will try to install 4.8.4 soon and I will try to analyze the error.

src/chain/block.cpp
```diff
@@ -138,43 +136,43 @@ bool block::operator!=(const block& other) const
 //-----------------------------------------------------------------------------
 
 // static
-block block::factory(const data_chunk& data, bool witness)
+block block::factory(const data_chunk& data, const bool witness)
```
evoskuil: As a matter of style we never use `const` on non-ref parameters. Since these are always copies it does not affect the caller, but limits internals, which is not necessary.

ldz1: @evoskuil: Ok, I get it. I will remove it.

However, there are different views. For that reason, I would suggest adding your statement into the style guide. 
https://github.com/libbitcoin/libbitcoin-system/wiki/Style-Guide

evoskuil: Will do, didn’t realize it wasn’t in there.

src/chain/block.cpp
```diff
@@ -120,7 +118,7 @@ block& block::operator=(block&& other)
     non_coinbase_inputs_ = other.non_coinbase_inputs_cache();
     header_ = std::move(other.header_);
     transactions_ = std::move(other.transactions_);
-    metadata = std::move(other.metadata);
+    metadata = other.metadata;
```
evoskuil: Please clarify why assignment is preferred.

src/chain/block.cpp
evoskuil: I just updated the style guide and added a couple more missing items.

src/chain/block.cpp
ldz1: >As a matter of style we never use const on non-ref parameters.

Done.

>Please clarify why assignment is preferred.

Because metadata's type is trivially-copyable and the std::move has no effect. That is the reason I decided to remove std::move. In addition, I've added a new unit test to see if the type stays trivially-copyable. 

The question is whether to use it preventively if the implementation of metadata's type changes. Moreover, one can argue in the following way, std::move belongs in the body of a move assignment operator, no matter if it brings something, to show the explicitly intent.

Similar question:
https://stackoverflow.com/questions/49241201/is-it-undesirable-to-defensively-apply-stdmove-to-trivially-copyable-types#

evoskuil: Thanks, that's what I figured, just wanted to make sure I wasn't missing something.

Clearly we wouldn't use `std::move` in the case of a integral data type that always copied, as that is self-documenting and guaranteed. But in the case of other types (which are always subject to internal change) the simplest approach is to always use `std::move` explicitly. This serves to make the code self-documenting (no need to search for the external guard, such as a unit test), and eliminates the additional cost of implementing and maintaining the external guard.

Also I think this is a pattern implemented throughout the codebase, so it's a question of consistent style as well.

ldz1:
Done.

src/chain/block.cpp
ldz1: Done.

src/chain/block.cpp
```diff
@@ -859,16 +857,16 @@ code block::accept(const chain_state& state,
...
+    const auto bip34 = state.is_enabled(bip34_rule);
+    const auto bip113 = state.is_enabled(bip113_rule);
+    const auto bip141 = state.is_enabled(bip141_rule);
 
     const auto max_sigops = bip141 ? max_fast_sigops : max_block_sigops;
     const auto block_time = bip113 ? state.median_time_past() :
         header_.timestamp();
 
-    if (header && (ec = header_.accept(state)))
+    if (header && ((ec = header_.accept(state))))
```
evoskuil: Good catch. I generally try to avoid this idiom except in the simple case of a single condition.

PR #1118, opened 2019-06-08
Add payment address parsing and tests for standard witness scripts

### LINKED ISSUE #844: Add payment address parsing for standard witness scripts.
This will allow the server to index payments accordingly.

thecodefactory:
Add payment address parsing and tests for standard witness scripts, including address formats specified in BIP 142 and BIP 173 (v0 only).

Resolves #844

### DIFF
--- a/include/bitcoin/system/wallet/witness_address.hpp
+++ b/include/bitcoin/system/wallet/witness_address.hpp
@@ -27,80 +27,85 @@ namespace system {
 namespace wallet {
 
 static BC_CONSTEXPR uint8_t padding = 0x00;
-static BC_CONSTEXPR size_t witness_p2wpkh_size = 3u + short_hash_size + checksum_size;
-static BC_CONSTEXPR size_t witness_p2wsh_size = 3u + hash_size + checksum_size;
-typedef byte_array<witness_p2wpkh_size> witness_p2wpkh;
+static BC_CONSTEXPR size_t witness_prefix_size = 3u;
+static BC_CONSTEXPR size_t witness_p2wsh_size = witness_prefix_size +
+    hash_size + checksum_size;
+static BC_CONSTEXPR size_t witness_p2wpkh_size = witness_prefix_size +
+    short_hash_size + checksum_size;
+
 typedef byte_array<witness_p2wsh_size> witness_p2wsh;
+typedef byte_array<witness_p2wpkh_size> witness_p2wpkh;
 
 /// A class for working with standard witness payment addresses.
 class BC_API witness_address
   : public payment_address
 {
-public:
-    enum class encoding: uint8_t
-    {
-        // BIP 142 p2wpkh and p2wsh address versions.
-        // https://github.com/bitcoin/bips/blob/master/bip-0142.mediawiki#specification
-        mainnet_base58_p2wpkh = 0x06,
-        testnet_base58_p2wpkh = 0x03,
-        mainnet_base58_p2wsh = 0x0a,
-        testnet_base58_p2wsh = 0x28,
-
-        // These encoding values are not from a specification.
-        mainnet_p2wpkh,
-        mainnet_p2wsh,
-        testnet_p2wpkh,
-        testnet_p2wsh,
-
-        unknown
-    };
-
+  public:
     static const std::string mainnet_prefix;
     static const std::string testnet_prefix;
+
+    static const std::string mainnet_p2wpkh;
+    static const std::string testnet_p2wpkh;
+
+    static const std::string mainnet_p2wsh;
+    static const std::string testnet_p2wsh;
+
+    static const uint8_t mainnet_bech32_p2wpkh;
+    static const uint8_t mainnet_bech32_p2wsh;
+
+    static const uint8_t testnet_bech32_p2wpkh;
+    static const uint8_t testnet_bech32_p2wsh;
+
+    static const uint8_t mainnet_base58_p2wpkh;
+    static const uint8_t mainnet_base58_p2wsh;
+
+    static const uint8_t testnet_base58_p2wpkh;
+    static const uint8_t testnet_base58_p2wsh;
 
     typedef std::vector<witness_address> list;
     typedef std::shared_ptr<witness_address> ptr;
 
     /// Create base58 witness programs using provided information.
-    static void to_witness(witness_p2wpkh& out, encoding out_type,
-        uint8_t witness_version, short_hash hash);
-    static void to_witness(witness_p2wsh& out, encoding out_type,
+    static void to_witness(witness_p2wsh& out, uint8_t version,
         uint8_t witness_version, hash_digest hash);
+    static void to_witness(witness_p2wpkh& out, uint8_t version,
+        uint8_t witness_version, short_hash hash);
 
     /// Constructors.
     witness_address();
     witness_address(witness_address&& other);
     witness_address(const witness_address& other);
     witness_address(const witness_p2wpkh& decoded,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh);
     witness_address(const witness_p2wsh& decoded,
-        encoding out_type=encoding::unknown);
-    witness_address(const data_chunk& data,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh);
     witness_address(const std::string& address,
-        encoding out_type=encoding::unknown);
-    witness_address(short_hash&& hash, encoding out_type=encoding::unknown,
-        uint8_t witness_version=0);
-    witness_address(const short_hash& hash, encoding out_type=encoding::unknown,
-        uint8_t witness_version=0);
-    witness_address(hash_digest&& hash, encoding out_type=encoding::unknown,
-        uint8_t witness_version=0);
+        uint8_t version=mainnet_bech32_p2wpkh);
+    witness_address(short_hash&& hash, uint8_t version=mainnet_bech32_p2wpkh,
+        uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
+    witness_address(const short_hash& hash,
+        uint8_t version=mainnet_bech32_p2wpkh, uint8_t witness_version=0,
+        const std::string& prefix=mainnet_prefix);
+    witness_address(hash_digest&& hash, uint8_t version=mainnet_bech32_p2wpkh,
+        uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
     witness_address(const hash_digest& hash,
-        encoding out_type=encoding::unknown,
-        uint8_t witness_version=0);
+        uint8_t version=mainnet_bech32_p2wpkh, uint8_t witness_version=0,
+        const std::string& prefix=mainnet_prefix);
     witness_address(const chain::script& script,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
     witness_address(const ec_private& secret,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
     witness_address(const ec_public& point,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
 
     /// Operators.
     bool operator<(const witness_address& other) const;
     bool operator==(const witness_address& other) const;
     bool operator!=(const witness_address& other) const;
     witness_address& operator=(const witness_address& other);
-    friend std::istream& operator>>(std::istream& in, witness_address& to);
     friend std::ostream& operator<<(std::ostream& out,
         const witness_address& of);
 
@@ -108,7 +113,7 @@ public:
     operator bool() const;
 
     /// Serializer.
-    std::string bech32(const std::string& prefix=mainnet_prefix) const;
+    std::string bech32() const;
     std::string encoded() const;
 
     /// Accessors.
@@ -127,21 +132,23 @@ private:
 
     /// Factories.
     static witness_address from_string(const std::string& address,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
     static witness_address from_witness(const witness_p2wpkh& decoded,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_base58_p2wpkh);
     static witness_address from_witness(const witness_p2wsh& decoded,
-        encoding out_type=encoding::unknown);
-    static witness_address from_data(const data_chunk& data,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_base58_p2wsh);
     static witness_address from_script(const chain::script& script,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
     static witness_address from_private(const ec_private& secret,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
     static witness_address from_public(const ec_public& point,
-        encoding out_type=encoding::unknown);
+        uint8_t version=mainnet_bech32_p2wpkh,
+        const std::string& prefix=mainnet_prefix);
 
-    encoding encoding_;
+    std::string prefix_;
     uint8_t witness_version_;
     hash_digest witness_hash_;
 };
--- a/include/bitcoin/system/wallet/witness_address.hpp
+++ b/include/bitcoin/system/wallet/witness_address.hpp
@@ -51,7 +51,9 @@ public:
         mainnet_p2wsh,
         testnet_p2wpkh,
         testnet_p2wsh,
-     };
+
+        unknown
+    };
 
     static const uint8_t mainnet_p2sh_p2wpkh;
     static const uint8_t mainnet_p2sh_p2wsh;
@@ -80,31 +82,28 @@ public:
     witness_address(witness_address&& other);
     witness_address(const witness_address& other);
     witness_address(const witness_p2wpkh& decoded,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     witness_address(const witness_p2wsh& decoded,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     witness_address(const data_chunk& data,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     witness_address(const std::string& address,
-        encoding out_type=encoding::mainnet_p2wpkh);
-    witness_address(short_hash&& hash,
-        encoding out_type=encoding::mainnet_p2wpkh,
+        encoding out_type=encoding::unknown);
+    witness_address(short_hash&& hash, encoding out_type=encoding::unknown,
         uint8_t witness_version=0);
-    witness_address(const short_hash& hash,
-        encoding out_type=encoding::mainnet_p2wpkh,
+    witness_address(const short_hash& hash, encoding out_type=encoding::unknown,
         uint8_t witness_version=0);
-    witness_address(hash_digest&& hash,
-        encoding out_type=encoding::mainnet_p2wpkh,
+    witness_address(hash_digest&& hash, encoding out_type=encoding::unknown,
         uint8_t witness_version=0);
     witness_address(const hash_digest& hash,
-        encoding out_type=encoding::mainnet_p2wpkh,
+        encoding out_type=encoding::unknown,
         uint8_t witness_version=0);
     witness_address(const chain::script& script,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     witness_address(const ec_private& secret,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     witness_address(const ec_public& point,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
 
     /// Operators.
     bool operator<(const witness_address& other) const;
@@ -133,19 +132,19 @@ private:
 
     /// Factories.
     static witness_address from_string(const std::string& address,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_witness(const witness_p2wpkh& decoded,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_witness(const witness_p2wsh& decoded,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_data(const data_chunk& data,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_script(const chain::script& script,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_private(const ec_private& secret,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
     static witness_address from_public(const ec_public& point,
-        encoding out_type=encoding::mainnet_p2wpkh);
+        encoding out_type=encoding::unknown);
 
     encoding encoding_;
     uint8_t witness_version_;
--- a/include/bitcoin/system/wallet/witness_address.hpp
+++ b/include/bitcoin/system/wallet/witness_address.hpp
@@ -33,8 +33,8 @@ class BC_API witness_address
   public:
     enum class address_format: uint8_t
     {
-        p2wpkh,
-        p2wsh
+        witness_pubkey_hash,
+        witness_script_hash
     };
 
     static const std::string mainnet_prefix;
@@ -48,27 +48,27 @@ class BC_API witness_address
     witness_address(witness_address&& other);
     witness_address(const witness_address& other);
     witness_address(const std::string& address,
-        address_format format=address_format::p2wpkh);
+        address_format format=address_format::witness_pubkey_hash);
     witness_address(short_hash&& hash,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
     witness_address(const short_hash& hash,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
     witness_address(hash_digest&& hash,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
     witness_address(const hash_digest& hash,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         uint8_t witness_version=0, const std::string& prefix=mainnet_prefix);
     witness_address(const chain::script& script,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
     witness_address(const ec_private& secret,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
     witness_address(const ec_public& point,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
 
     /// Operators.
@@ -76,6 +76,8 @@ class BC_API witness_address
     bool operator==(const witness_address& other) const;
     bool operator!=(const witness_address& other) const;
     witness_address& operator=(const witness_address& other);
+    friend std::istream& operator>>(std::istream& in,
+        witness_address& to);
     friend std::ostream& operator<<(std::ostream& out,
         const witness_address& of);
 
@@ -101,16 +103,16 @@ private:
 
     /// Factories.
     static witness_address from_string(const std::string& address,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
     static witness_address from_script(const chain::script& script,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
     static witness_address from_private(const ec_private& secret,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
     static witness_address from_public(const ec_public& point,
-        address_format format=address_format::p2wpkh,
+        address_format format=address_format::witness_pubkey_hash,
         const std::string& prefix=mainnet_prefix);
 
     std::string prefix_;
--- a/include/bitcoin/system/wallet/witness_address.hpp
+++ b/include/bitcoin/system/wallet/witness_address.hpp
@@ -67,9 +67,9 @@ class BC_API witness_address
 
     /// Create base58 witness programs using provided information.
     static void to_witness(witness_p2wsh& out, uint8_t version,
-        uint8_t witness_version, hash_digest hash);
+        uint8_t witness_version, const hash_digest& hash);
     static void to_witness(witness_p2wpkh& out, uint8_t version,
-        uint8_t witness_version, short_hash hash);
+        uint8_t witness_version, const short_hash& hash);
 
     /// Constructors.
     witness_address();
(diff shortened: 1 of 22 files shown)


thecodefactory:
Interesting build error (fails on clang, good on gcc).  Testing clang locally to resolve.

thecodefactory:
I think this is in decent shape for merging (in terms of correctness), although I expect the interface may change some after @evoskuil has a chance to simplify it.

I have a separate (incoming, but nearly complete) PR for explorer that really makes this usable (including ec-to-witness and witness-to-key commands).

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,175 @@
...
+    static const uint8_t testnet_p2sh_p2wpkh;
+    static const uint8_t testnet_p2sh_p2wsh;
+
+    static const std::string mainnet_prefix;
+    static const std::string testnet_prefix;
+
+    typedef std::vector<witness_address> list;
+    typedef std::shared_ptr<witness_address> ptr;
+
+    /// Public for unit testing.
```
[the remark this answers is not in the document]
thecodefactory: Ah, good to know, thanks!

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,544 @@
...
+    encoding out_type)
+{
+    // Attempt to decode BIP 173 address format.
+    base32 bech32_decoded;
+    if (decode_base32(bech32_decoded, address))
+    {
+        const uint8_t witness_version = bech32_decoded.payload.front();
+
+        // Checks specific to witness version 0.
+        if (witness_version == 0 && (address.size() < 42 || address.size() > 62))
```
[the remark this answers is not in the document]
thecodefactory: :+1:  Agreed, these BIP specified lengths should be better handled.

include/bitcoin/system/wallet/witness_address.hpp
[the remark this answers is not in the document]
evoskuil: This sort of thing is really a pain in the ass in C++. I miss the use of [moq](https://github.com/moq/moq) and [friend assemblies](https://docs.microsoft.com/en-us/dotnet/standard/assembly/friend-assemblies?tabs=csharp), no need to expose internals just to properly unit test.

src/wallet/witness_address.cpp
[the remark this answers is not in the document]
evoskuil: Yes, IAW style guide, give each magic number a name.

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,169 @@
...
+#include <bitcoin/system/math/hash.hpp>
+#include <bitcoin/system/wallet/payment_address.hpp>
+
+namespace libbitcoin {
+namespace system {
+namespace wallet {
+
+static BC_CONSTEXPR uint8_t padding = 0x00;
+static BC_CONSTEXPR size_t witness_p2wpkh_size = 3u + short_hash_size + checksum_size;
+static BC_CONSTEXPR size_t witness_p2wsh_size = 3u + hash_size + checksum_size;
```
evoskuil: Declare `witness_prefix_size = 3u`. Should update other classes as well so that magic numbers are self-documenting.

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,169 @@
...
+        testnet_base58_p2wsh = 0x28,
+
+        // These encoding values are not from a specification.
+        mainnet_p2wpkh,
+        mainnet_p2wsh,
+        testnet_p2wpkh,
+        testnet_p2wsh,
+
+        unknown
+    };
```
evoskuil: For consistency you should expose the "base58" values as to `static const` and as named in `payment_address()`. You should drop the `enum` as the remaining are redundant and reduces flexibility. Undeclared prefixes must be allowed for alts.

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,169 @@
...
+    witness_address(const witness_address& other);
+    witness_address(const witness_p2wpkh& decoded,
+        encoding out_type=encoding::unknown);
+    witness_address(const witness_p2wsh& decoded,
+        encoding out_type=encoding::unknown);
+    witness_address(const data_chunk& data,
+        encoding out_type=encoding::unknown);
+    witness_address(const std::string& address,
+        encoding out_type=encoding::unknown);
+    witness_address(short_hash&& hash, encoding out_type=encoding::unknown,
```
evoskuil: Avoid using the `out_` prefix for an in parameter. Just name the parameter `version` as it matches your `enum` comment and `payment_address`.

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,169 @@
...
+    witness_address& operator=(const witness_address& other);
+    friend std::istream& operator>>(std::istream& in, witness_address& to);
+    friend std::ostream& operator<<(std::ostream& out,
+        const witness_address& of);
+
+    /// Cast operators.
+    operator bool() const;
+
+    /// Serializer.
+    std::string bech32(const std::string& prefix=mainnet_prefix) const;
```
evoskuil: This method should not be provided as it allows one to change the prefix for an already constructed address. The address object should contain all information upon construct. This also implies there is no use for the `unknown` version value.

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,169 @@
...
+        uint8_t witness_version, short_hash hash);
+    static void to_witness(witness_p2wsh& out, encoding out_type,
+        uint8_t witness_version, hash_digest hash);
+
+    /// Constructors.
+    witness_address();
+    witness_address(witness_address&& other);
+    witness_address(const witness_address& other);
+    witness_address(const witness_p2wpkh& decoded,
+        encoding out_type=encoding::unknown);
```
evoskuil: An address object should never have an unknown version value. The object encompasses a complete address.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+
+BOOST_AUTO_TEST_CASE(base32__witness_address_vectors__valid_expected)
+{
+    for (const auto& test: witness_address_tests)
+    {
+        base32 decoded;
+        BOOST_REQUIRE(decode_base32(decoded, test.address));
+        uint8_t witness_version = decoded.payload.front();
+        // https://github.com/sipa/bech32/blob/master/ref/javascript/tests.js#L24
+        if (witness_version != 0)
```
evoskuil: Avoid conditional tests (test style guide).

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+            witness_version += 0x50;
+
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(5, 8, false,
```
evoskuil: Avoid wrapping test lines (test style guide)

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+    // Ensure they encode equally.
+    BOOST_REQUIRE_EQUAL(address.bech32("tb"), address2.bech32("tb"));
+    BOOST_REQUIRE_EQUAL(address.encoded(), address2.encoded());
+}
+
+BOOST_AUTO_TEST_CASE(witness_address__witness_vectors__valid_expected)
+{
+    for (const auto& test: native_witness_tests)
+    {
+        if (test.encoding == witness_address::encoding::mainnet_base58_p2wpkh ||
```
evoskuil: Avoid conditional tests.

test/wallet/witness_address.cpp
evoskuil: This is an address, not a converter, so just refer to it as `address`.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(5, 8, false,
+            decoded.payload, 1);
+        BOOST_REQUIRE(witness_program.size() > 1 &&
+                      witness_program.size() < 41);
+        const auto witness_program_size = static_cast<uint8_t>(
+            witness_program.size());
+
+        data_chunk result = { witness_version, witness_program_size };
+        extend_data(result, witness_program);
```
evoskuil: I think there's an overload for extending using a set of values, so this could be simplified.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+        // https://github.com/sipa/bech32/blob/master/ref/javascript/tests.js#L24
+        if (witness_version != 0)
+            witness_version += 0x50;
+
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(5, 8, false,
+            decoded.payload, 1);
+        BOOST_REQUIRE(witness_program.size() > 1 &&
+                      witness_program.size() < 41);
+        const auto witness_program_size = static_cast<uint8_t>(
```
evoskuil: Move up so and use in test macros so this is not computed 3 times.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+// https://github.com/bitcoin/bips/blob/master/bip-0173.mediawiki#test-vectors
+struct test_address
+{
+    std::string address;
+    std::string decoded;
+};
+
+typedef std::vector<test_address> test_address_list;
+
+test_address_list witness_address_tests =
```
evoskuil: Use `const`.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+{
+    // witness address, decoded
+    { "BC1QW508D6QEJXTDG4Y5R3ZARVARY0C5XW7KV8F3T4", "0014751e76e8199196d454941c45d1b3a323f1433bd6" },
+    { "tb1qrp33g0q5c5txsp9arysrx4k6zdkfs4nce4xj0gdcccefvpysxf3q0sl5k7", "00201863143c14c5166804bd19203356da136c985678cd4d27a1b8c6329604903262" },
+    { "bc1pw508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7k7grplx", "5128751e76e8199196d454941c45d1b3a323f1433bd6751e76e8199196d454941c45d1b3a323f1433bd6" },
+    { "BC1SW50QA3JX3S", "6002751e" },
+    { "bc1zw508d6qejxtdg4y5r3zarvaryvg6kdaj", "5210751e76e8199196d454941c45d1b3a323" },
+    { "tb1qqqqqp399et2xygdj5xreqhjjvcmzhxw4aywxecjdzew6hylgvsesrxh6hy", "0020000000c4a5cad46221b2a187905e5266362b99d5e91c6ce24d165dab93e86433" }
+};
```
evoskuil: As a constant declaration these can be nested (style).

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+        uint8_t witness_version = decoded.payload.front();
+        // https://github.com/sipa/bech32/blob/master/ref/javascript/tests.js#L24
+        if (witness_version != 0)
+            witness_version += 0x50;
+
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(5, 8, false,
+            decoded.payload, 1);
+        BOOST_REQUIRE(witness_program.size() > 1 &&
+                      witness_program.size() < 41);
```
evoskuil: Avoid magic number literals.

test/wallet/witness_address.cpp
evoskuil: Avoid magic number literals.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+BOOST_AUTO_TEST_CASE(base32__witness_address_vectors__valid_expected)
+{
+    for (const auto& test: witness_address_tests)
+    {
+        base32 decoded;
+        BOOST_REQUIRE(decode_base32(decoded, test.address));
+        uint8_t witness_version = decoded.payload.front();
+        // https://github.com/sipa/bech32/blob/master/ref/javascript/tests.js#L24
+        if (witness_version != 0)
+            witness_version += 0x50;
```
evoskuil: Avoid magic number literals.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+}
+
+struct witness_address_accessor
+  : public witness_address
+{
+    data_chunk convert_bits(uint32_t from_bits, uint32_t to_bits,
+        bool pad, const data_chunk& in, size_t in_offset)
+    {
+        return witness_address::convert_bits(
+            from_bits, to_bits, pad, in, in_offset);
```
evoskuil: Normally want to break as close as possible to 80 chars, vs. nest args.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,213 @@
...
+struct test_vector
+{
+    std::string hash;
+    std::string address;
+    bool testnet;
+    witness_address::encoding encoding;
+    uint8_t witness_version;
+};
+
+typedef std::vector<test_vector> test_list;
```
evoskuil: No need for a `typedef` here (or even an explicit type?).

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,632 @@
...
+}
+
+witness_address::witness_address(const witness_p2wsh& decoded,
+    encoding out_type)
+  : witness_address(from_witness(decoded, out_type))
+{
+}
+
+witness_address::witness_address(const data_chunk& data, encoding out_type)
+  : witness_address(from_data(data, out_type))
```
evoskuil: Why is arbitrary data allowed?

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,632 @@
...
+const size_t bech32_contracted_bit_size = 5;
+const size_t bech32_expanded_bit_size = 8;
+
+// Discovery based on string is not ideal.  This use case is for
+// accepting a witness_address as a string without a further specified
+// encoding type.
+static witness_address::encoding get_bech32_encoding(const std::string& address,
+    size_t hash_length)
+{
+    const auto prefix = address.substr(0, 2);
```
evoskuil: Safe for < 2 char string?

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,632 @@
...
+    if (prefix == "bc")
+        return hash_length == short_hash_size ?
+            witness_address::encoding::mainnet_p2wpkh :
+            witness_address::encoding::mainnet_p2wsh;
+    else if (prefix == "tb")
+        return hash_length == short_hash_size ?
+            witness_address::encoding::testnet_p2wpkh :
+            witness_address::encoding::testnet_p2wsh;
```
evoskuil: How would this work for alts?

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,632 @@
...
+// encoding type.
+static witness_address::encoding get_bech32_encoding(const std::string& address,
+    size_t hash_length)
+{
+    const auto prefix = address.substr(0, 2);
+    if (prefix == "bc")
+        return hash_length == short_hash_size ?
+            witness_address::encoding::mainnet_p2wpkh :
+            witness_address::encoding::mainnet_p2wsh;
+    else if (prefix == "tb")
```
evoskuil: Avoid magic literals.

src/wallet/witness_address.cpp
evoskuil: Avoid magic literals.

src/formats/base_32.cpp
```diff
@@ -122,8 +122,8 @@ data_chunk checksum(const base32& value)
     data_chunk checksum(checksum_size);
     const auto modified = polymod(expanded) ^ 1;
 
-    for (size_t index = 0; index < checksum.size(); ++index)
-        checksum[index] = (modified >> (5 * (5 - index))) & 31;
+    for (size_t index = 0; index < checksum_size; ++index)
```
evoskuil: Use `auto` (no need for more than int here).

src/formats/base_32.cpp
```diff
@@ -122,8 +122,8 @@ data_chunk checksum(const base32& value)
...
+        checksum[index] = (modified >> 5 * (5 - index)) & 31;
```
evoskuil: Avoid magic literals.

src/formats/base_32.cpp
thecodefactory: After considering, I left this (and a different instance, below) as-is to avoid compiler warnings:

```
src/formats/base_32.cpp: In function ‘uint32_t libbitcoin::system::polymod(const data_chunk&)’:
src/formats/base_32.cpp:102:36: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
         for (auto index = 0; index < bit_group_size; ++index)
                              ~~~~~~^~~~~~~~~~~~~~~~
src/formats/base_32.cpp: In function ‘libbitcoin::system::data_chunk libbitcoin::system::checksum(const libbitcoin::system::base32&)’:
src/formats/base_32.cpp:127:32: warning: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]
     for (auto index = 0; index < checksum_size; ++index)
```
Alternatively, could use int for the constants ...

thecodefactory:
Updated, but I need to track down the clang presented bugs next ...

thecodefactory:
Updated.

@evoskuil The use of `version` is all around much better than the previous encoding stuff when it's well defined and used properly.  This part still seems problematic to me:

https://github.com/libbitcoin/libbitcoin-system/pull/1118/files#diff-e73f757b65a6660336e35268a78db663R48

In short, I'm using some `fake` version codes simply to distinguish the cases where a witness address is constructed from a script or public key and we need to distinguish between the identity as bech32 p2wpkh or p2wsh encoding.  It could probably be replaced with something as simple as an additional `bool` on construction, but wanted your input on it.  `Version` is currently re-used in this case because it's already there (and mimics the actually defined versions for the base58 equivalents), but as it is now, it does mean that we maintain fake version mappings that may eventually conflict with actual versions (perhaps in the alt space).  If we did add a bool for that determination, the `version` is ignored for cases where we don't use the base58 encoding.  The other downside of not using the `version` mapping is that it makes more conditional logic required for things like this:

https://github.com/libbitcoin/libbitcoin-system/pull/1118/files#diff-e73f757b65a6660336e35268a78db663R299

thecodefactory:
Just pushed one more minor update/clean-up ...

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,550 @@
...
+            return {};
+    }
+}
+
+// Methods.
+// ----------------------------------------------------------------------------
+
+/// static
+void witness_address::to_witness(witness_p2wpkh& out, uint8_t version,
+    uint8_t witness_version, short_hash hash)
```
evoskuil: Avoid parameter copies by always using `const short_hash& hash` idiom). When a copy is required, make it internally for the sake of consistency.

evoskuil: See other instances.

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,550 @@
...
+{
+    return encoded() < other.encoded();
+}
+
+bool witness_address::operator==(const witness_address& other) const
+{
+    return valid_ == other.valid_ && version_ == other.version_ &&
+        prefix_ == other.prefix_ &&
+        witness_version_ == other.witness_version_ && hash_ == other.hash_ &&
+        witness_hash_ == other.witness_hash_;
```
evoskuil: This is the case where I prefer stacking (a long sequence of `&&` or `||` operations), an exception to the stacking style rule. 
```cpp
    return valid_ == other.valid_
        && version_ == other.version_
        && prefix_ == other.prefix_
        && witness_version_ == other.witness_version_
        && hash_ == other.hash_
        && witness_hash_ == other.witness_hash_;
```

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,265 @@
...
+{
+    { "bc1pw508d6qejxtdg4y5r3zarvary0c5xw7kw508d6qejxtdg4y5r3zarvary0c5xw7k7grplx", "5128751e76e8199196d454941c45d1b3a323f1433bd6751e76e8199196d454941c45d1b3a323f1433bd6" },
+    { "BC1SW50QA3JX3S", "6002751e" },
+    { "bc1zw508d6qejxtdg4y5r3zarvaryvg6kdaj", "5210751e76e8199196d454941c45d1b3a323" },
+};
+
+BOOST_AUTO_TEST_CASE(witness_address__construct__from_ec_public__valid_expected)
+{
+    const witness_address address(ec_public(BIP142_PUBLIC_KEY),
+        witness_address::mainnet_base58_p2wpkh);
```
evoskuil: Avoid test line wrap, lots more below. Can use intermediate vars to break up (perf not a material consideration).

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,265 @@
...
+    // between the defined OP_0 opcode (value 0) and the OP_1 opcode
+    // (value 0x51)
+    const auto op_version_gap = 0x50;
+    const auto witness_v0_program_minimum_limit = 1;
+    const auto witness_v0_program_maximum_limit = 41;
+
+    for (const auto& test: witness_address_nonzero_version_tests)
+    {
+        base32 decoded;
+        BOOST_REQUIRE(decode_base32(decoded, test.address));
```
evoskuil: I tend to group the statements such that `BOOST_REQUIRE(...)` (or a set of them) terminates the section, leaving a blank line after, then on to the next group.

evoskuil:
> I'm using some fake version codes simply to distinguish the cases where a witness address is constructed from a script or public key and we need to distinguish between the identity as bech32 p2wpkh or p2wsh encoding.

Okay, so there is an enumeration of possible methods of encoding, that is independent of the other information, but currently it is being inferred from that information? This calls for an independent enumeration, but as you suggest, since there are only two valid options as per the specification, it could be Boolean. I would refer to it as something other than "version" or "encoding", given that those are overloaded in the same context. The current distinction is between "public-key" and "script" style of hash/address. Maybe a class enumeration named `address_type` with two options, `key` and `script`. Allowing an invalid option, such as `unknown` will only unnecessarily complicate the code, given that the class enum is a constraint.

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,550 @@
...
+}
+
+// static
+witness_address witness_address::from_private(const ec_private& secret,
+    uint8_t version, const std::string& prefix)
+{
+    if (secret)
+        return { secret.to_public(), version, prefix };
+
+    return {};
```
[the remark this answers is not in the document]
thecodefactory: For almost all of the ternary return cases, we get this error if attempted:

```
error: initializer list cannot be used on the right hand side of operator '?'
...
error: initializer list cannot be used on the right hand side of operator ':'
```

It's probably not worth first assigning, then using the ternary return, as it doesn't simplify/shorten things.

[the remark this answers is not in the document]
evoskuil: You need to explicitly specify the type and it will work.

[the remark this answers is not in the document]
evoskuil: Not a huge deal tho.

thecodefactory:
>> I'm using some fake version codes simply to distinguish the cases where a witness address is constructed from a script or public key and we need to distinguish between the identity as bech32 p2wpkh or p2wsh encoding.

> Okay, so there is an enumeration of possible methods of "encoding", that is independent of the other information, but currently it is being inferred from that information? This calls for an independent enumeration, but as you suggest, since there are only two valid options as per the specification, it could be Boolean. I would refer to it as something other than "version" or "encoding", given that those are overloaded in the same context. The current distinction is between "public-key" and "script" style of hash/address. Maybe a class enumeration named address_format with two options, public_key and script. Allowing an invalid option, such as unknown will only unnecessarily complicate the code, given that the class enum is a constraint. Then using format for the parameter names.

Sort of.  When the witness address is constructed from hash/public/key/script, we want to know what will come out when `.encoded()` is called (for example, base58 p2wpkh, base58 p2wsh, bech32 p2wpkh, or bech32 p2wsh).  Construction takes that data, a version (required in the base58 cases to distinguish between p2wpkh and p2wsh, and re-used from the `payment_address` base), a witness version, and a prefix (unused for base58, but required for bech32).  What we know with those parameters for base58 encodings is everything required, and we have an extra unused prefix.  What we don't explicitly know with those parameters is whether the encoding should be bech32 p2wpkh, or bech32 p2wsh.  That's where we're re-using the version field (since it's only used in the base58 encoding cases otherwise).  We could add a new field, that simply specifies bech32 p2wpkh or bech32 p2wsh (and/or possibly mandates the use of base58 instead).

IFF we did that, in the base58 cases, now we ignore the prefix and the new encoding flag (unused).  And similarly in the bech32 cases, we ignore the version parameter passed in (only used to specify a base58 version).  This gets a little tricky because now we don't explicitly know which to ignore (the base58 version, or the new encoding flag, perhaps without some other piece of information).

Does this make sense?  Re-using the version (as we do now) appears to be the most concise, but it's confusing.

At least conceptually, what we may want is to split this out some more so that we have separate classes (naming is bad, just explicit for example):

```
base58_witness_address(hash/script/public_key, version, witness_version);

bech32_witness_address(hash/script/public_key, witness_version,
    prefix, p2wpkh_or_p2wsh_flag);
```

_Where hash/script/public_key means we'd have constructors for each of those (in addition to the other standard variants)_

The alternative is cramming it all into something like:

```
witness_address(hash/script/public_key, version, witness_version,
    prefix, p2wpkh_or_p2wsh_flag);
```

Which is what happens if we add a new flag/format field.

evoskuil:
Why are we attempting to support distinct base58 and bech32 encodings?

thecodefactory:
> Why are we attempting to support distinct base58 and bech32 encodings?

This implementation supports both BIP142 ("Address Format for Segregated Witness") and BIP173 ("Base32 address format for native v0-16 witness outputs").

evoskuil:
Overloading leads to complexity.

thecodefactory:
> Overloading leads to complexity.

Agreed, thus the design issue I'm fretting over :-(

But I also think splitting them out completely (like the poorly named examples above) is not quite right either.  I'm open to ideas/suggestions here.  I know there's a very workable middle ground somewhere, but I'm not seeing it just yet.

evoskuil:
[BIP142](https://github.com/bitcoin/bips/blob/master/bip-0142.mediawiki) is status: withdrawn.

thecodefactory:
> [BIP142](https://github.com/bitcoin/bips/blob/master/bip-0142.mediawiki) is status: withdrawn.

Wow, thanks, didn't realize!  If the answer is as simple as stripping out BIP142 support, I think I've got a clear resolution path here.

thecodefactory:
Updated, but will remove BIP142 support next to help simplify the interface.

thecodefactory:
Dropped BIP142 support and added an `address_format` field.  I think this is much less confusing than the previous code trying to support both base58 and bech32 addresses all in one.

evoskuil:
`magic_numbers[index]` - creative yet ineffective. The point of naming magic numbers is to eliminate the magic. This is still magic. ;)

src/chain/script.cpp
```diff
@@ -1171,6 +1177,24 @@ operation::list script::to_pay_multisig_pattern(uint8_t signatures,
     return ops;
 }
 
+operation::list script::to_pay_witness_key_hash_pattern(const short_hash& hash)
+{
+    return operation::list
```
evoskuil: `operation::list` may be dropped here as it's the return type (more below).

include/bitcoin/system/wallet/witness_address.hpp
```diff
@@ -0,0 +1,142 @@
...
+
+/// A class for working with standard witness payment addresses.
+class BC_API witness_address
+  : public payment_address
+{
+  public:
+    enum class address_format: uint8_t
+    {
+        p2wpkh,
+        p2wsh
```
evoskuil: Style: acronyms :/. Could you review against existing similar terms in `-system` and update for consistency?

src/formats/base_32.cpp
```diff
@@ -91,13 +93,15 @@ uint32_t polymod(const data_chunk& values)
         0x3b6a57b2, 0x26508e6d, 0x1ea119fa, 0x3d4233dd, 0x2a1462b3
```
evoskuil: The magic numbers have moved. But what are these?

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,367 @@
...
+std::string witness_address::encoded() const
+{
+    base32 bech32;
+    bech32.prefix = prefix_;
+    bech32.payload = build_chunk(
+    {
+        to_chunk(witness_version_),
+        witness_hash_ == null_hash ? convert_bits(bech32_expanded_bit_size,
+            bech32_contracted_bit_size, true, to_chunk(hash_), 0) :
+                convert_bits(bech32_expanded_bit_size,
```
evoskuil: You could outdent this "else" portion one tab width.

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,367 @@
...
+const hash_digest& witness_address::witness_hash() const
+{
+    return witness_hash_;
+}
+
+chain::script witness_address::output_script() const
+{
+    return format_ == address_format::p2wpkh ?
+        chain::script::to_pay_witness_key_hash_pattern(hash_) :
+        chain::script::to_pay_witness_script_hash_pattern(witness_hash_);
```
evoskuil: Notice the above-mentioned outdenting is done here (same logical level).

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,180 @@
...
+        uint8_t witness_version = decoded.payload.front();
+        BOOST_REQUIRE(witness_version == 0);
+
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(
+            bech32_contracted_bit_size, bech32_expanded_bit_size, false,
+            decoded.payload, 1);
+
+        const auto witness_program_size = static_cast<uint8_t>(witness_program.size());
+        BOOST_REQUIRE(witness_program_size > witness_v0_program_minimum_limit &&
```
evoskuil: Test style: unwrap or (better) split into two tests.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,180 @@
...
+    for (const auto& test: witness_address_nonzero_version_tests)
+    {
+        base32 decoded;
+        BOOST_REQUIRE(decode_base32(decoded, test.address));
+        uint8_t witness_version = decoded.payload.front();
+        BOOST_REQUIRE(witness_version != 0);
+        witness_version += op_version_gap;
+
+        witness_address_accessor converter;
+        const auto witness_program = converter.convert_bits(
```
evoskuil: Test style: unwrap.

test/wallet/witness_address.cpp
```diff
@@ -0,0 +1,180 @@
...
+            bech32_contracted_bit_size, bech32_expanded_bit_size, false,
+            decoded.payload, 1);
+
+        const auto witness_program_size = static_cast<uint8_t>(witness_program.size());
+        BOOST_REQUIRE(witness_program_size > witness_v0_program_minimum_limit &&
```
evoskuil: Test style: unwrap or split.

thecodefactory:
@evoskuil I believe this has been updated to address all current feedback.  If I missed something, let me know!

src/formats/base_32.cpp
```diff
@@ -86,22 +88,32 @@ data_chunk expand(const std::string& prefix)
 // Do the checksum math.
 uint32_t polymod(const data_chunk& values)
 {
-    static const uint32_t magic_numbers[] =
+    // Polynomials in the bech32 implementation are represented by
+    // simple integers. Generally 30-bit integers are used, where each
+    // bit corresponds to one coefficient of the polynomial.
+    //
+    // https://bitcoin.stackexchange.com/questions/74573/how-is-bech32-based-on-bch-codes
```
evoskuil: Drop "https://" to shorten line and eliminate hyperlink highlighting in VS (which is annoying).

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,383 @@
...
+    auto hash = to_array<short_hash_size>(converted);
+    return { std::move(hash), format, witness_version, prefix };
+}
+
+// static
+witness_address witness_address::from_private(const ec_private& secret,
+    address_format format, const std::string& prefix)
+{
+    return secret ? witness_address(secret.to_public(), format, prefix)
+        : witness_address();
```
evoskuil: witness_address{}

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,383 @@
...
+        return {};
+
+    data_chunk data;
+    if (!point.to_data(data))
+        return {};
+
+    const uint8_t witness_version = 0;
+
+    return format == address_format::witness_pubkey_hash ?
+        witness_address(bitcoin_short_hash(data), format, witness_version, prefix) :
```
evoskuil: line length, maybe just use `version`.

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,383 @@
...
+    base32 bech32;
+    bech32.prefix = prefix_;
+    bech32.payload = build_chunk(
+    {
+        to_chunk(witness_version_),
+        witness_hash_ == null_hash ?
+            convert_bits(bech32_expanded_bit_size, bech32_contracted_bit_size,
+                true, to_chunk(hash_), 0) :
+            convert_bits(bech32_expanded_bit_size, bech32_contracted_bit_size,
+                true, to_chunk(witness_hash_), 0)
```
evoskuil: Use named const for `0` parameter.

src/wallet/witness_address.cpp
```diff
@@ -0,0 +1,383 @@
...
+
+    data_chunk out;
+    for (const auto value: in)
+    {
+        if (offset++ < in_offset)
+            continue;
+
+        accumulator = ((accumulator << from_bits) | value) &
+            maximum_accumulator;
+        bit_count += from_bits;
```
evoskuil: Could move this up one line for slightly better appearance.

PR #1126, opened 2019-08-14
Movable scalar type

narodnik:
Major changes:
1. Movable ec_scalar type. It uses an internal pointer via std::unique_ptr which means memory is no longer contiguous and could cause contention via pointer indirection, but the benefit of being able to use rvalue's in complicated expressions without copying outweighs the cost of 1) not being able to pre-allocate large contiguous areas of memory 2) co-locate values on the stack.
2. Constructor by value which is explicit.
3. Filled in missing operators == and !- as well as copy/move operators (following rule of 5).


narodnik:
Note: std::array is a non-movable type

test/math/ec_scalar.cpp
```diff
@@ -50,6 +50,24 @@ BOOST_AUTO_TEST_CASE(ec_scalar__add__valid__expected)
     BOOST_REQUIRE_EQUAL(encode_base16(sum.secret()), SUM1);
 }
 
+BOOST_AUTO_TEST_CASE(ec_scalar__from__uint64__valid__expected)
```
evoskuil: Pattern: class__method__test_vector__expected_result
Recommend: ec_scalar__sum__from_uint64__expected

test/math/ec_scalar.cpp
```diff
@@ -50,6 +50,24 @@ BOOST_AUTO_TEST_CASE(ec_scalar__add__valid__expected)
...
+{
+    const ec_scalar x(110);
+    const ec_scalar y(256);
+    const auto sum = x + y;
+    BOOST_REQUIRE(sum);
+    const ec_scalar result(110 + 256);
```
evoskuil: All things being equal I recommend use of “expected” as the expectation var.

test/math/ec_scalar.cpp
```diff
@@ -50,6 +50,24 @@ BOOST_AUTO_TEST_CASE(ec_scalar__add__valid__expected)
...
+    BOOST_REQUIRE_EQUAL(sum, result);
+}
+
+BOOST_AUTO_TEST_CASE(ec_scalar__move__success)
```
evoskuil: Suggest: ec_scalar__construct__move__initialized
Another test: ec_scalar__construct__move__value_transferred
(I would also test that the expected value has been moved, vs. just object initialized/deinitialized.)

src/math/ec_scalar.cpp
```diff
@@ -18,36 +18,82 @@
...
 // ----------------------------------------------------------------------------
 
+ec_scalar& ec_scalar::operator=(uint64_t value)
+{
+    secret_.reset(new ec_secret);
+    BITCOIN_ASSERT(secret_->size() == 32);
+    auto serial = bc::system::make_unsafe_serializer(secret_->begin());
+    serial.write_8_bytes_big_endian(0);
+    serial.write_8_bytes_big_endian(0);
+    serial.write_8_bytes_big_endian(0);
```
evoskuil: Is there a 16 byte writer?

src/wallet/ec_private.cpp
```diff
@@ -160,18 +160,21 @@ ec_private ec_private::from_uncompressed(const wif_uncompressed& wif,
 // Conversion to WIF loses payment address version info.
 std::string ec_private::encoded() const
 {
+    if (!secret_)
+        return {};
```
evoskuil: Good

src/math/ec_scalar.cpp
evoskuil: You can probably use partial aggregate initialization here to fill the zeroes. You could also use it for the decoded int, but prob simpler to use the writer for that.

evoskuil: You could also use our zeroize function to full the zeroes. Much faster than writing, maybe not faster than aggregate init.

include/bitcoin/system/math/ec_scalar.hpp
```diff
@@ -52,11 +57,17 @@ class BC_API ec_scalar
...
+    friend ec_scalar operator+(ec_scalar left, const ec_scalar& right);
+    friend ec_scalar operator-(ec_scalar left, const ec_scalar& right);
+    friend ec_scalar operator*(ec_scalar left, const ec_scalar& right);
+
+    using ec_secret_uniq = std::unique_ptr<ec_secret>;
+
+    ec_secret_uniq secret_;
 };
 
+bool operator==(const ec_scalar& left, const ec_scalar& right);
```
evoskuil: Cannot these all be member operators? Better organizationally to have members than free-floating functions, even for operator overrides, where possible. This is a common pattern throughout the libs.

narodnik:
We can solve the problem with allocating contiguous memory by providing a constructor which accepts a pointer to a memory location.

narodnik:
Actually the proper way is to provide an allocator with instantiation of the class (like `std::vector<Allocator = ...>`).

evoskuil:
We should avoid exposing internals.

include/bitcoin/system/math/ec_scalar.hpp
```diff
@@ -32,17 +33,21 @@ class BC_API ec_scalar
 
     /// Constructors.
     ec_scalar();
+    explicit ec_scalar(uint64_t value);
```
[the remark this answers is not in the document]
narodnik: No, I prefer it without explicit. I will remove it.

include/bitcoin/system/math/ec_scalar.hpp
```diff
@@ -32,17 +33,21 @@ class BC_API ec_scalar
...
     ec_scalar(const ec_secret& secret);
+    ec_scalar(const ec_scalar& scalar);
+    ec_scalar(ec_scalar&& scalar);
+    ~ec_scalar() = default;
```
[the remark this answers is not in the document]
narodnik: This is explicitly provided because of the Rule of 5. We have copy constructor move constructor, copy operator and move operator. Ro5 states there should also be a destructor. So in this case, I instantiate the default one.

src/math/ec_scalar.cpp
```diff
@@ -18,36 +18,82 @@
...
+    *this = std::move(scalar);
 }
 
 // Operators.
 // ----------------------------------------------------------------------------
 
+ec_scalar& ec_scalar::operator=(uint64_t value)
+{
+    secret_.reset(new ec_secret);
+    BITCOIN_ASSERT(secret_->size() == 32);
```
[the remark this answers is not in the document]
narodnik: Done.

src/math/ec_scalar.cpp
```diff
@@ -18,36 +18,82 @@
...
+{
+    *this = std::move(scalar);
 }
 
 // Operators.
 // ----------------------------------------------------------------------------
 
+ec_scalar& ec_scalar::operator=(uint64_t value)
+{
+    secret_.reset(new ec_secret);
```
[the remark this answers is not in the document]
narodnik: Unfortunately `std::make_unique` is not C++11 only 14 or later, otherwise I prefer to use that. See https://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique

include/bitcoin/system/math/ec_scalar.hpp
```diff
@@ -52,11 +57,17 @@ class BC_API ec_scalar
...
 
 protected:
-    // These should be const, apart from the need to implement assignment.
-    bool valid_;
-    ec_secret secret_;
+    friend ec_scalar operator+(ec_scalar left, const ec_scalar& right);
+    friend ec_scalar operator-(ec_scalar left, const ec_scalar& right);
+    friend ec_scalar operator*(ec_scalar left, const ec_scalar& right);
+
+    using ec_secret_uniq = std::unique_ptr<ec_secret>;
```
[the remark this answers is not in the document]
narodnik: I think this policy should be changed. The consensus/recommendation in C++ now is to prefer using over typedef. About the abbreviation, should I use `_uptr` instead?

src/math/ec_scalar.cpp
```diff
@@ -71,19 +117,46 @@ ec_scalar& ec_scalar::operator-=(const ec_scalar& scalar)
...
 
-ec_scalar& ec_scalar::operator=(const ec_secret& secret)
+ec_scalar::operator bool() const
 {
-    valid_ = true;
-    secret_ = secret;
-    return *this;
+    // Caller must make both test against zero when that is required.
+    // Overloading bool with valid and non-zero would be very non-intuitive.
+    return static_cast<bool>(secret_);
```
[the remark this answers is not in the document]
narodnik: Yes it requires an explicit cast. `std::unique_ptr::operator bool` is marked as explicit, so it does not cast implicitly. (unfortunately)

src/math/ec_scalar.cpp
narodnik: Fixed with `std::fill`

include/bitcoin/system/math/ec_scalar.hpp
narodnik: I think the common argument is that free floating operators are better than member ones because then LHS can do implicit conversions from other types like `4 == ec_scalar(4)`. Whereas member operators cannot do this conversion so that would be an invalid statement with a member operator.

test/math/ec_scalar.cpp
narodnik: Done

test/math/ec_scalar.cpp
narodnik: Done

test/math/ec_scalar.cpp
narodnik: Done & Done

include/bitcoin/system/math/ec_scalar.hpp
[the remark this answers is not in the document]
narodnik: Done

src/math/ec_scalar.cpp
narodnik: Done

include/bitcoin/system/math/ec_scalar.hpp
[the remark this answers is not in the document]
narodnik: Check out page 63 "Item 9: Prefer alias declarations to typedefs" in the Scott Meyers "Effective Modern C++" book for a good explanation of this topic and why using is preferable. I even saw somewhere there is a discussion about later deprecating typedefs in C++ in the long term.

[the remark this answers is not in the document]
narodnik: Changed type from `_uniq` to `_uptr` prefix.

narodnik:
Yes it's just the old design (which is non-movable) might have some other performance advantages (contiguous memory for some algorithms sometimes might improve performance because the CPU cache remains hot)... But this is not an optimization that I need currently (maybe later).

include/bitcoin/system/math/ec_scalar.hpp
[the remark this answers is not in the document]
evoskuil: Why would it be necessary to make it explicit to comply with this guideline? Not having an explicit destructor implies a default destructor. We don't do this in any class, so I would opt for consistency over adherence to a distinction without a difference.

src/math/ec_scalar.cpp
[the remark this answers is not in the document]
evoskuil: Ah, thanks. We don't use unique pointer anywhere. It always seems preferable to use shared pointer and explicitly dump the pointer where desired. As it's strictly a class internal this is always possible and adds no more code. I would generally avoid its use in any public interface as well.

include/bitcoin/system/math/ec_scalar.hpp
```diff
@@ -61,9 +61,9 @@ class BC_API ec_scalar
     friend ec_scalar operator-(ec_scalar left, const ec_scalar& right);
     friend ec_scalar operator*(ec_scalar left, const ec_scalar& right);
 
-    using ec_secret_uniq = std::unique_ptr<ec_secret>;
+    using ec_secret_uptr = std::unique_ptr<ec_secret>;
```
evoskuil: Given that `using` and `typedef` are functionally equivalent in aliasing, I don’t have a preference other than consistency. We have avoided `using` (except in one instance where there is no alternative) in class definitions (and headers generally, as free aliasing in headers leads to insidious namespace conflicts). We avoid `typedef` in implementation, preferring `using` in that context. We could change that at some point, but all things being equal consistency is the goal.

evoskuil: I see no need for an alias here as it is never used except to declare the field.

evoskuil: Ad-hoc acronyms are also avoided (along with abbreviations). We consistently use `ptr` for smart pointers where the distinction is important. As this is the only non-shared smart pointer in the codebase we could standardize on `uptr`, though I’d like to avoid proliferation of type identifiers in variable naming (see the Hungarian mess). So I’d prob just opt for `ptr` here. But for reasons previously mentioned, I think this would be better implemented as a shared pointer.

src/math/ec_scalar.cpp
```diff
@@ -60,11 +60,12 @@ ec_scalar::ec_scalar(ec_scalar&& scalar)
 ec_scalar& ec_scalar::operator=(uint64_t value)
 {
     secret_.reset(new ec_secret);
-    BITCOIN_ASSERT(secret_->size() == 32);
-    auto serial = bc::system::make_unsafe_serializer(secret_->begin());
-    serial.write_8_bytes_big_endian(0);
-    serial.write_8_bytes_big_endian(0);
-    serial.write_8_bytes_big_endian(0);
+    static_assert(ec_secret_size == 32, "invalid size for ec_secret");
+    auto last_int_iterator = secret_->end() - 8;
```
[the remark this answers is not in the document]
evoskuil: Prefer `std::prev()’ to offset iterator.

PR #1128, opened 2019-08-16
Remove duplicate code in timer

narodnik:
This code is not needed, since the std::forward optimizes out the intervening function call and the result is exactly similar.


evoskuil:
Thanks!

PR #1129, opened 2019-08-18
remove instructions about old version of gcc 4.8

narodnik:
Current version of gcc is 9.3... even Debian stable which has very old software uses gcc 6.3 now. I think it's safe to remove these instructions now.


evoskuil:
These exist in at least one other repo.

PR #1131, opened 2019-08-21
Regenerate build artifacts.

thecodefactory:


thecodefactory:
@evoskuil If this succeeds and we're ok moving to clang-5.0 from clang-3.4 (clang 9 is current/stable, but many distros continue to have back to the 3.x series), I'll push to all repos and merge.

PR #1142, opened 2019-09-27
Chain metadata and filter containing class for blockchain.

pmienk:


pmienk:
Any suggestions for naming? Didn't want to use the message class names as none contain both filter and filter header.

PR #1143, opened 2019-10-21
compact filter message corrections, computation signature change.

pmienk:


include/bitcoin/system/utility/neutrino_filter.hpp
```diff
@@ -37,7 +37,7 @@ namespace libbitcoin {
 namespace system {
 namespace neutrino {
 
-data_chunk compute_filter(const chain::block& validated_block);
+bool compute_filter(const chain::block& validated_block, data_chunk& out_filter);
```
evoskuil: Out parameters should be to the left of the parameter list.

src/utility/neutrino_filter.cpp
```diff
@@ -33,12 +33,11 @@ namespace libbitcoin {
 namespace system {
 namespace neutrino {
 
-data_chunk compute_filter(const chain::block& validated_block)
+bool compute_filter(const chain::block& validated_block, data_chunk& out_filter)
```
evoskuil: Out parameters should be to the left of the parameter list.

Issue #1149, opened 2019-10-28
Appveyor build failure, string too big.


evoskuil:
https://ci.appveyor.com/project/pmienk/libbitcoin-system/builds/28421815/job/2538rwfvcgajb7q1#L365
```
neutrino_filter.cpp
..\..\..\..\test\utility\neutrino_filter.cpp(213): error C2026: string too big, trailing characters truncated [c:\git\libbitcoin-system\builds\msvc\vs2013\libbitcoin-system-test\libbitcoin-system-test.vcxproj]
..\..\..\..\test\utility\neutrino_filter.cpp(234): error C2026: string too big, trailing characters truncated [c:\git\libbitcoin-system\builds\msvc\vs2013\libbitcoin-system-test\libbitcoin-system-test.vcxproj]
```

pmienk:
Tests disabled on msvc.

PR #1150, opened 2019-10-28
Fix windows build failure on test case.

pmienk:


evoskuil:
https://github.com/libbitcoin/libbitcoin-system/issues/1149

evoskuil:
```
..\..\..\..\test\utility\neutrino_filter.cpp(2399): fatal error C1091: compiler limit: string exceeds 65535 bytes in length
```

test/utility/neutrino_filter.cpp
```diff
@@ -208,9 +208,1209 @@ BOOST_AUTO_TEST_CASE(compute_filter_header__block_1414221__success)
     BOOST_REQUIRE_EQUAL(result, expected);
 }
 
+#ifndef _MSC_VER
```
evoskuil: Please add comment that explains the exclusion(s).

PR #1151, opened 2019-10-29
Implement bip42 soft fork (finite monetary supply).

evoskuil:
This will require a database resync as the fork enumeration has changed.


kulpreet:
Interesting. This PR is essentially allowing users to disable bip42, as otherwise it was enabled by default. Right?

evoskuil:
Yes, all forks are configurable, we just missed this one in config.

Issue #1180, opened 2021-02-16
Add pbkdf2_sha256 function to the public API of libbitcoin-system


ghallak:
Is there a way for the library users to call the function `pbkdf2_sha256`? I've noticed that it's available in the external code here https://github.com/libbitcoin/libbitcoin-system/blob/master/src/math/external/pbkdf2_sha256.h but there is no function in the public API that calls this function.

Would it be fine to add a function to https://github.com/libbitcoin/libbitcoin-system/blob/master/include/bitcoin/system/math/hash.hpp that calls `pbkdf2_sha256`?

thecodefactory:
Interesting.  It is used internally in the `scrypt_hash` method, but appears to not be exported.  Adding it should be fine, but please add appropriate tests vectors in https://github.com/libbitcoin/libbitcoin-system/blob/master/test/math/hash.cpp for correctness.

PR #1181, opened 2021-02-16
Add pbkdf2_hmac_sha256 function to the public API

ghallak:
Fixes: #1180


thecodefactory (approved):
Looks good, and thanks for the tests.  If there are well known test vectors, it would be good to make sure that those are in (but I didn't check that these aren't).

PR #1216, opened 2021-03-25
Fixes electrum mnemonic validation by requiring the initial entropy.

thecodefactory:
Validation is now as follows: Provided a word list and a starting entropy value, it is possible to create a new mnemonic, which should match the provided word list.

There are a couple electrum test cases that do not provide the original entropy, so we don't validate those.  Many more electrum test cases do provide the entropy, so those are now properly validated.

### DIFF
--- a/src/wallet/electrum.cpp
+++ b/src/wallet/electrum.cpp
@@ -484,7 +484,7 @@ static cpp_int mnemonic_decode(const word_list& mnemonic,
         if (position == -1)
             return { 1 };
 
-        entropy *= dictionary_size + position;
+        entropy = (entropy * dictionary_size) + position;
     }
 
     return entropy;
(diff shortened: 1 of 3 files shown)


src/wallet/electrum.cpp
```diff
@@ -533,18 +534,29 @@ long_hash decode_mnemonic(const word_list& mnemonic,
...
+    //
+    // Creating a mnemonic is self validating, in that we guarantee
+    // that the entropy used to create it decodes back to that same
+    // entropy value.  Validation ensures that the passed in mnemonic
+    // and entropy is consistent by verifying a match on re-creation.
+    const auto created_mnemonic = create_mnemonic(entropy, lexicon, prefix);
+
     return is_new_seed(mnemonic, get_seed_prefix(prefix)) &&
-        (mnemonic_decode(mnemonic, lexicon) == 0);
+        (mnemonic_decode(mnemonic, lexicon) == mnemonic_decode(created_mnemonic, lexicon));
```
thecodefactory, line 551: Note that we can't just do this here:

```C++
 return is_new_seed(mnemonic, get_seed_prefix(prefix)) &&
        (mnemonic_decode(mnemonic, lexicon) == entropy);
```

The reason is because the starting entropy does not have to equal the final entropy.  The electrum algorithm advances it in order to create the mnemonic, so the best way to validate it seems to tie it to a newly generated one as this PR does.

src/wallet/electrum.cpp
```diff
@@ -484,7 +484,7 @@ static cpp_int mnemonic_decode(const word_list& mnemonic,
         if (position == -1)
             return { 1 };
 
-        entropy *= dictionary_size + position;
+        entropy = (entropy * dictionary_size) + position;
```
evoskuil, line 487: I thought that this might have been the intent. Almost made this change, but obv that felt sketchy.

PR #1219, opened 2021-03-28
Consolidate dictionaries, match full Electrum and bip39 lists, rename seed.

evoskuil:


evoskuil:
Yes

mxaddict:
> Yes

Was it related to performance during compilation or just general clean up?

evoskuil:
It pertains to the desired const and compile time construction of a single data structure that overlays the memory of the object code.

mxaddict:
I see, thanks 👍🏼

evoskuil:
No problem, see also: https://github.com/libbitcoin/libbitcoin-system/blob/master/include/bitcoin/system/impl/words/dictionary.ipp#L35-L40

PR #1235, opened 2021-10-05
Enable builds with c++20 and boost 1.76

pool2win:
I am planning on using `system` with braidpool, locally I am using boost 1.76 and gcc/c++20.

To be sure `system` worked with those, I tried to run this locally and had to make a few changes for the tests to pass. These changes are backwards compatible.

Required only very few changes, mostly in unicode tests.

### DIFF
--- a/test/unicode/unicode.cpp
+++ b/test/unicode/unicode.cpp
@@ -493,7 +493,9 @@ BOOST_AUTO_TEST_CASE(unicode__to_utf16_array__ascii__expected)
     uint8_t truncated;
     const auto size = to_utf16(truncated, utf16, sizeof(utf16), utf8.c_str(), (int)utf8.size());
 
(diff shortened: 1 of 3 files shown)


test/unicode/unicode.cpp
```diff
@@ -511,7 +513,9 @@ BOOST_AUTO_TEST_CASE(unicode__to_utf16_array__non_ascii__expected)
     uint8_t truncated;
     const auto size = to_utf16(truncated, utf16, sizeof(utf16), utf8.c_str(), (int)utf8.size());
 
-    BOOST_REQUIRE_EQUAL(utf16, expected_utf16.c_str());
```
evoskuil, line 514: Maybe:
```
BOOST_REQUIRE(std::wstring(utf16), expected_utf16);
```

evoskuil:
This should be overcome by pending master PR and CI integration.

PR #1236, opened 2021-10-12
Add github actions for CI

pool2win:
Add support for github actions to run tests. 

Results from the latest run are here: https://github.com/pool2win/libbitcoin-system/actions/runs/1332074390. This run used cached depdencies, and the build time is around 17m.

Currently works for all the same combinations on Travis, apart from macOs dynamic link libs.

### Caching

The icu, boost builds are cached using github actions' "cache action". This results in reuse of the built dependencies for a branch and all the children branches.

### Runs for

The github CI action will run for PR's to `master` branch and for pushes to `ci/github-actions` branch - for debugging the ci setup.

### Next tasks

Need to figure out how to make the macOS shenanigans with rpath work. I can run things locally by using `install_name_tool`, but I need help to move it into libbitcoin-build or somewhere else where it can be automated for CI and user setups.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,136 @@
+name: Run tests
+
+on:
+  push:
+    branches:
+      - $default-branch
+      - 'ci/github-actions'
+  pull_request:
+    branches: [ $default-branch ]
+
+jobs:
+  build:
+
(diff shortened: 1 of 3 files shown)


evoskuil:
This is being integrated into -build by @pmienk

PR #1286, opened 2022-12-21
Setup tx for signing using connect

kulpreet:
This is a draft PR for initial discussion on how to continue on the work to provide transaction signing using connect.

## What we currently do 

1. Added a new method in transaction called
   `transaction::generate_signatures` that calls a new method in
   `interpreter` calls `interpreter::connect_for_signing`.
2. The `interpreter::connect_for_signing` calls the
   `interpreter::connect` creates a `program` with `sign_mode_`
   boolean flag set to true.
3. If this flag is set during a programs `run`, then we sign the
   transaction and gather signatures.
4. To gather signatures, we track generated signatures in a vector of
   `der_signature`s in the `program`.
5. For signing, we provide the private keys in the signature
   fields. We append sighash flags to the private key. So the
   signature in sign mode has `<private_key><flags>`.
6. Once signing is completed by the `check_sig_verify` and
   `checksig_multisig_verify` methods the interpreter takes the
   signatures from the programs and adds them to the out param
   `signatures` passed in to the `connect_for_signing` and
   `transaction::connect` methods.

image

## What we should consider doing

1. We should use template member function to evaluate the `if
   (sign_mode_)` condition at compile time. This will require turning
   `connect` and `run` into template member function that take a bool
   template argument to identify if things are running in signing mode or not.
2. We should refactor `interpreter::connect` to break it into smaller methods
   1. It might be worthwhile to consider a new class like
      `transaction_connector` that hides away all the connection logic
      in one place away from the `transaction` class.
   2. This will help a lot in working on templatizing the `connect`
      member functions.
3. We also need to write tests to cover the segwit cases.

### DIFF
--- a/builds/cmake/CMakeLists.txt
+++ b/builds/cmake/CMakeLists.txt
@@ -515,6 +515,7 @@ if (with-tests)
         "../../test/chain/script.hpp"
         "../../test/chain/stripper.cpp"
         "../../test/chain/transaction.cpp"
+        "../../test/chain/transaction_sign.cpp"
         "../../test/chain/witness.cpp"
         "../../test/chain/enums/opcode.cpp"
         "../../test/config/base16.cpp"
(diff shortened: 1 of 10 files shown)


kulpreet:
Closing this for now.

PR #1292, opened 2023-01-05
Refactor connecting p2sh into a separate method

pool2win:
This PR moves code that handles connecting p2sh into a separate function.

Let me know if these changes potentially break anything, I can add new tests to be sure we don't break anything.

### DIFF
--- a/include/bitcoin/system/impl/machine/interpreter.ipp
+++ b/include/bitcoin/system/impl/machine/interpreter.ipp
@@ -1644,9 +1644,9 @@ connect(const context& state, const transaction& tx,
             return ec;
     }
 
+    // Witness must be empty if no bip141 or invalid witness program (bip141).
     else if (!witnessed && !input.witness().stack().empty())
     {
-        // Witness must be empty if no bip141 or invalid witness program (bip141).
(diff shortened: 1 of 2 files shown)


evoskuil:
Please rebase.

pool2win:
> Please rebase.

Done.

evoskuil (approved):
The refactor makes methods smaller but also creates an additional stack call and input dereferences, but reasonable tradeoff for the improved readability.

include/bitcoin/system/impl/machine/interpreter.ipp
```diff
@@ -1640,35 +1640,54 @@ connect(const context& state, const transaction& tx,
...
 
-        if (!program.is_true(false))
-            return error::stack_false;
+template <typename Stack>
+code interpreter<Stack>::connect_p2sh(const context& state,
+    const transaction& tx, const input_iterator& it,
+    interpreter& input_program) NOEXCEPT
+{
+    code ec;
+    const auto input = **it;
```
evoskuil, line 1662: This is a copy of the input, should be:
`const auto input = **it;`

PR #1294, opened 2023-01-09
Initial work for signing transactions with prepare

pool2win:
This PR ports over the signing mode on top of refactoring work to separate `connect_witness` and `connect_embeded`.

This should still not be merged as this brings in allocation and condition jumps in the transaction verification.

We can avoid at least the allocation by either using polymorphism for `program` and template functions for `connect`. I'll have a go at this.

This PR is just to share the current state and get feedback.


evoskuil:
Putting this on ice for now.

PR #1302, opened 2023-01-16
Demo: Fix memory owner issue pointed by clang-tidy

pool2win:


evoskuil (commented):
Clang is pointing to an issue that is explicitly ack’d by the warning suppressions. This avoids unnecessary object and pointer move/copy operations.

PR #1306, opened 2023-01-25
Use explicit instantiation for accumulator<sha256>

pool2win:
On Linux with gcc, we get libbitcoin-system.so => 61MB and libbitcoin-system-test => 117MB.

Removing implicit instantiation for a single template instance reduces the above to 49MB and 111MB.

### CI Failures

- I need to update libbitcoin-build to add the new file ../../src/hash/accumulators.cpp to CMakeLists.txt, autotools and vs config? 
- I only the new file to CMakeLists.txt for my local setup, so only cmake builds will pass.


evoskuil:
Always run libbitcoin-build codegen when adding, removing, moving, or renaming source files.

evoskuil:
https://github.com/libbitcoin/libbitcoin-system/pull/1307

PR #1502, opened 2024-07-18
Regenerate artifacts with SSE4.1 correction.

pmienk:


parsevalbtc:
Thanks @pmienk. Now optimizations can fully compile on Linux.

PR #1535, opened 2024-10-22
Fix: Remove redundant error code argument from std::filesystem::exist…

gubatron:
- Removed the second argument (`error_code`) from the std::filesystem::exists call in `parser.cpp`, as the function does not accept an error code parameter in C++17 and later.
  - This resolves a compatibility issue with modern C++ standards.

### DIFF
--- a/src/config/parser.cpp
+++ b/src/config/parser.cpp
@@ -94,7 +94,7 @@ bool parser::load_configuration_variables(variables_map& variables,
 
     // If the existence test errors out we pretend there's no file :/.
     error_code code;
-    if (!config_path.empty() && exists(config_path, code))
+    if (!config_path.empty() && exists(config_path))
     {
         const auto& path = config_path.string();
         ifstream file(path);


gubatron:
This was breaking the build for me on macos building with `std=c++20`

```
$ clang --version
Apple clang version 16.0.0 (clang-1600.0.26.3)
Target: arm64-apple-darwin23.5.0
Thread model: posix
```

evoskuil:
Thanks, so you have a successful ARM build? We don’t CI that platform yet.

evoskuil:
Documentation states that the API is new in C++17, and the change to the #2 override causes a change from `noexcept` to throws.

gubatron:
ran again configure with `C++17` on macos arm, this happens, both here and in `libbitcoin-network`:

```
...
checking pkg-config is at least version 0.9.0... yes
checking whether g++ supports C++20 features with -std=c++20... no
checking whether g++ supports C++20 features with +std=c++20... no
checking whether g++ supports C++20 features with -h std=c++20... no
configure: error: *** A compiler with support for C++20 language features is required.
```

evoskuil:
C++20 is required, but the API is “since C++17” which includes C++20. Great to have you onboard!

evoskuil:
Also, we would really love to have an macOS ARM build set up in our GitHub Actions CI process!

evoskuil:
See: https://github.com/libbitcoin/libbitcoin-node/issues/686

PR #1586, opened 2025-01-20
Specify std::error_code usage in parser.cpp, without it fails to comp…

Deltajom:
…ile on CPP17 amd above

### DIFF
--- a/src/config/parser.cpp
+++ b/src/config/parser.cpp
@@ -93,7 +93,7 @@ bool parser::load_configuration_variables(variables_map& variables,
     const auto config_path = get_config_option(variables, option_name);
 
     // If the existence test errors out we pretend there's no file :/.
-    error_code code;
+    std::error_code code;
     if (!config_path.empty() && exists(config_path, code))
     {
         const auto& path = config_path.string();


src/config/parser.cpp
```diff
@@ -93,7 +93,7 @@ bool parser::load_configuration_variables(variables_map& variables,
     const auto config_path = get_config_option(variables, option_name);
 
     // If the existence test errors out we pretend there's no file :/.
-    error_code code;
```
evoskuil, line 96: This is a `boost::system::error_code` , please provide compilation error details.

evoskuil (changes requested):

src/config/parser.cpp
Deltajom: When specified to boost::system::error_code only instead of std::error_code compilation of libbitcoin fails due to CPP compiler (CPP20 for me) defaulting to use the std::filesystem::exists, meaning std::error_code solved my issue

Other solution is to specify boost::filesystem::exists and  boost::system::error_code, can try this further as well

Depends on what is best for the rest of project, have attached compilation error in full to this reply as well.
[fullerror.txt](https://github.com/user-attachments/files/18471831/fullerror.txt)

evoskuil: Ok, makes sense. This is an artifact of out change from `boost::filesystem` to `std::filesystem` .

Please also remove `using namespace boost::system;`

evoskuil:
Thanks!

Issue #1623, opened 2025-03-30
Extended key lineage.depth overflows and produces garbage key


quapka:
I've tested the following on `version3` and also `master` (the code needs to be slightly different for the `version3` to compile.

```cpp
#include <bitcoin/system.hpp>
#include <string.h>
#include <iostream>

#include <stdio.h>
#include <stdlib.h>

using namespace bc;
using namespace libbitcoin::system;
using namespace bc::system::wallet;


int main() {

    // Example tests from https://github.com/bitcoin/bips/blob/master/bip-0032.mediawiki#test-vector-1
    data_chunk hd_seed = {0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x06, 0x07, 0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f};

    /* data_chunk seed_chunk(to_chunk(hd_seed)); */
    hd_private m(hd_seed, hd_private::mainnet);
    hd_public y = m.to_public();

    for (int i = 0; i < 255; i++) {
        y = y.derive_public(0);
    }
    printf("%d: ", y.lineage().depth);
    std::cout << y << std::endl;

    y = y.derive_public(0);
    printf("%d: ", y.lineage().depth);
    std::cout << y << std::endl;

    return 0;
}
```
(The example does not show this for `hd_private`, because I couldn't get the demo to derive 256th key, but the problem is most likely the same for `hd_private` as well.)

which gives the following output:
```text
255: xpubEND4cWBkwMUcwj3bjw4RNYcpnuvgbEaGSCAujB1XQro3Ptpvs8hDMFsBmk1mhfz9sGc3k4XPpueGAcR66Kb7HMXwfnKKBaV3i7YyMxLuwKh
0: 1111111111111111111111111111111111111111111111111111111111111111111111111111114rcJhr
```

For the `version3` case it seemed clear, why the overflow happens in [`hd_private.cpp`](https://github.com/libbitcoin/libbitcoin-system/blob/version3/src/wallet/hd_private.cpp#L259) and [`hd_public.cpp`](https://github.com/libbitcoin/libbitcoin-system/blob/version3/src/wallet/hd_public.cpp#L231). I haven't investigated further, where does the `1111111111111111111111111111111111111111111111111111111111111111111111111111114rcJhr` value come from though.

[...]

PR #1624, opened 2025-03-31
Test for `lineage.depth` overflow of the extended keys

quapka:
Hi, based on the discussion in https://github.com/libbitcoin/libbitcoin-system/issues/1623, I've decided to create a few test cases. I've manually created an extended private key with the depth set to `254` and then incremented it twice, checking that the first derivation passes and the second one fails (for various combinations of `hd_private`, `hd_public` and `derive_private` or `derive_public`).

I've split the tests into two commits due to, surprisingly, some variants failing. In particular, it seems to be possible to hit the overflow without and when deriving a public key of depth `256` from a private key of depth `255`. Unfortunately, the tests are quite some copy-paste (I am happy to be transparent about this), but I could not see where the test (for the failing ones) would be incorrect.


quapka:
If I read the output of the CI correctly, it runs all tests together as a single test as part of the installation in `install-cmake.sh` and thus the failure that I talked about is hidden :-/

evoskuil:
https://github.com/libbitcoin/libbitcoin-system/actions/runs/14169924138/job/39725696803?pr=1624#step:10:958

quapka:
> https://github.com/libbitcoin/libbitcoin-system/actions/runs/14169924138/job/39725696803?pr=1624#step:10:958
> 
> cmake: https://github.com/libbitcoin/libbitcoin-system/actions/runs/14169924138/job/39725696825#step:23:1

My bad, thanks! The two failingk tests are indeed the ones I was talking about. The offending lines is [this](https://github.com/libbitcoin/libbitcoin-system/pull/1624/commits/b290f4fdd4a1efd834f3eae94d3d69cf31a65cc9#diff-ab970f44c1adf16dec0fc1750b9525c93483181df17090bd29c21901c2334a59R195) and [this one](https://github.com/libbitcoin/libbitcoin-system/pull/1624/commits/b290f4fdd4a1efd834f3eae94d3d69cf31a65cc9#diff-ab970f44c1adf16dec0fc1750b9525c93483181df17090bd29c21901c2334a59R217). 

The idea of the tests was to hit the overflow when going from 255 xpriv to 256/0 xpub. [This](https://github.com/libbitcoin/libbitcoin-system/blob/master/src/wallet/keys/hd_private.cpp#L269) code.

Oh, wait a minute. Isn't the `derive_private(index)` return value left unchecked and simple the `to_public()` is what gets returned, which succeeds even for the empty key?

evoskuil:
No worries, in some of the CI outputs (CMake I think) you have to go to a different section to see the individual test failures.

quapka:
> Oh, wait a minute. Isn't the `derive_private(index)` return value left unchecked and simple the `to_public()` is what gets returned, which succeeds even for the empty key?

I am starting to believe that this is the case. The tests pass with the following patch:
```diff
diff --git a/src/wallet/keys/hd_private.cpp b/src/wallet/keys/hd_private.cpp
index fa9d7dc9c..0b4201c74 100644
--- a/src/wallet/keys/hd_private.cpp
+++ b/src/wallet/keys/hd_private.cpp
@@ -266,7 +266,12 @@ hd_private hd_private::derive_private(uint32_t index) const NOEXCEPT

 hd_public hd_private::derive_public(uint32_t index) const NOEXCEPT
 {
-    return derive_private(index).to_public();
+    const auto tmp = derive_private(index);
+    if (tmp) {
+        return tmp.to_public();
+    }
+
+    return {};
 }

 // Operators.

```

I am no C/C++, so the code is just something quickly stitched together. Wdyt?

evoskuil:
That makes sense, I'll give it a look in a bit.

quapka:
If my suggestion is indeed the problem, I wonder what a proper fix should be. Maybe verifying `this` within `to_public()` method? Which makes me wonder, how the C/C++ API of HD keys work and what are possibly all the other places, where operation are done on invalid keys.

evoskuil:
I think people generally ignore secp invalidity and guard against deriving past the limit (since it's a guaranteed failure).

quapka:
Do you also backport bug-fixes? I see the same thing in [`verson3`](https://github.com/libbitcoin/libbitcoin-system/blob/version3/src/wallet/hd_private.cpp#L269).

evoskuil:
Sure, if they're simple.

evoskuil:
It's either:

```cpp
hd_public hd_private::derive_public(uint32_t index) const NOEXCEPT
{
    const auto derived = derive_private(index);
    return derived ? derived.to_public() : hd_public{};
}
```
or:
```cpp
hd_public hd_private::to_public() const NOEXCEPT
{
    if (!valid_)
        return {};
        ...
 }
```
as you suggested. I prefer the latter because it explicitly guards both scenarios which also benefits the case where the caller has not guarded a private key before conversion (as opposed to derivation) to public.

evoskuil:
Feel free to make the change into both repos.

quapka:
> ```c++
> hd_public hd_private::to_public() const NOEXCEPT
> {
>     if (!valid_)
>         return {};
>         ...
>  }
> ```

Thanks for the code examples. I also like the second one more. I've added a couple more tests to also:
- test explicitly for decoding null keys, both `hd_private` and `hd_public`
- test `to_public()` called on invalid `hd_private` arising from other reasons, as you mentioned.

- [ ] Provide fix to `master`
- [ ] Provide fix to `version3`

src/wallet/keys/hd_private.cpp
```diff
@@ -231,6 +231,11 @@ hd_key hd_private::to_hd_key() const NOEXCEPT
 
 hd_public hd_private::to_public() const NOEXCEPT
 {
+
+    if (!valid_) {
+        return {};
+    }
```
evoskuil, line 237: Avoid bracing for single line (also braces would be balanced).

test/wallet/keys/hd_private.cpp
```diff
@@ -128,4 +128,112 @@ BOOST_AUTO_TEST_CASE(hd_private__derive_public__long_seed__expected)
     BOOST_REQUIRE_EQUAL(m0xH1yH2_pub.encoded(), "xpub6FnCn6nSzZAw5Tw7cgR9bi15UV96gLZhjDstkXXxvCLsUXBGXPdSnLFbdpq8p9HmGsApME5hQTZ3emM2rnY5agb9rXpVGyy3bdW6EEgAtqt");
 }
 
+BOOST_AUTO_TEST_CASE(hd_private__constructor__null_key_decodes_to_invalid__expected)
```
evoskuil, line 131: This:

`hd_private__constructor__null_key_decodes_to_invalid__expected`

would be better as:

`hd_private__constructor__null_key__decodes_to_invalid`

Others as well. Third part is the thing being tested and fourth part is the expectation.

test/wallet/keys/hd_public.cpp
```diff
@@ -93,4 +103,28 @@ BOOST_AUTO_TEST_CASE(hd_public__derive_public__long_seed__expected)
...
+    const auto xpub_255 = xpub_254.derive_public(1);
+    const auto xpub_256 = xpub_255.derive_public(0);
+
+    BOOST_REQUIRE_EQUAL(xpub_254.lineage().depth, 254);
+    BOOST_REQUIRE(xpub_254);
+    // the maximal valid depth is 255
+    BOOST_REQUIRE_EQUAL(xpub_255.lineage().depth, 255);
+    BOOST_REQUIRE(xpub_255);
+
+    // depth overflows uint from 255 to 0
```
evoskuil, line 123: I don't think it's actually overflowing here, it should be the default (invalid/null) key, and the depth value for that happens to be zero.

evoskuil (changes requested):
Could you please squash these to a single commit?

src/wallet/keys/hd_private.cpp
quapka: I like to be explicit about the `if` scope, as I have seen a new code added (maybe under the influence of Python-like, whitespace aware languages), without adding the braces. Tests should hopefully catch that though.

evoskuil: https://github.com/libbitcoin/libbitcoin-system/wiki/Style-Guide

evoskuil:
Moved to https://github.com/libbitcoin/libbitcoin-system/pull/1637

PR #1637, opened 2025-04-13
Guard deriving public from invalid extended private key, update tests.

evoskuil:
Rebased, style corrections, squashed commits.


quapka:
Many thanks for finalizing this!

evoskuil:
Thanks for submitting it!

PR #1652, opened 2025-05-08
Stub in taproot program version handling, rename enum values.

evoskuil:


include/bitcoin/system/chain/enums/script_version.hpp
```diff
@@ -28,11 +28,11 @@ namespace chain {
 /// Script versions (bip141).
 enum class script_version
 {
-    /// Defined by bip141. 
-    zero = 0,
+    /// If the version byte is 0 (either P2WPKH or P2WSH or inavalid) [bip141].
```
MarnixCroes, line 31: nit
```suggestion
    /// If the version byte is 0 (either P2WPKH or P2WSH or invalid) [bip141].
```

evoskuil:
Thanks!

include/bitcoin/system/chain/enums/script_version.hpp
evoskuil: Thanks! I think these had just been obsoleted in subsequent work.

evoskuil: I meant to commit this but merged the PR instead - from my phone :?

PR #1703, opened 2025-06-24
install scripts: add -h as alias for --help

MarnixCroes:
All executables in Libbitcoin have `--help` and `-h`, however `-h` is missing at the install scripts.

Which is unexpected and it happened to me a couple times that I used `-h` and that the install script executed instead of displaying help.


MarnixCroes:
if approved, I'll make the same PR for the other libraries

pmienk:
As these are generated, the change needs to be made within `libbitcoin-build`.

MarnixCroes:
closing in favor of https://github.com/libbitcoin/libbitcoin-system/pull/1704

PR #1713, opened 2025-08-09
Refactor IPv6/IPv4 handling and console echo logic

ryihan:
Updated IPv6 to IPv4 conversion to avoid deprecated Boost methods and implemented manual byte extraction. Refactored console echo functions to use termios on non-Windows platforms and added stubs for Windows, improving cross-platform compatibility.

### DIFF
--- a/src/unicode/utf8_everywhere/stdio.cpp
+++ b/src/unicode/utf8_everywhere/stdio.cpp
@@ -17,15 +17,12 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include <bitcoin/system/unicode/utf8_everywhere/stdio.hpp>
-#include <windows.h>
 #ifdef HAVE_MSC
     #include <fcntl.h>
     #include <io.h>
     #include <windows.h>
 #else
-   #if !defined(_WIN32)
-#include <termios.h>
-#endif
+    #include <termios.h>
 #endif
 #include <mutex>
 #include <bitcoin/system/define.hpp>
@@ -38,7 +35,6 @@ namespace system {
 
 LCOV_EXCL_START("Untestable but visually-verifiable section.")
 
-#ifdef HAVE_MSC
 
 // The width of utf16 stdio buffers.
 constexpr size_t utf16_buffer_size = 256;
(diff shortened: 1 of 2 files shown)


src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -17,13 +17,15 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include <bitcoin/system/unicode/utf8_everywhere/stdio.hpp>
-
+#include <windows.h>
```
evoskuil, line 20: This can only be included under `HAVE_MSC` (and is below).

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -17,13 +17,15 @@
...
 #ifdef HAVE_MSC
     #include <fcntl.h>
     #include <io.h>
     #include <windows.h>
 #else
-    #include <termios.h>
+   #if !defined(_WIN32)
```
evoskuil, line 26: This is a redundant condition (and native PP defines should be avoided).

evoskuil (changes requested):
Thanks for the submission! It would help to split these out into two independent PRs. Please see comments inline.

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -17,7 +17,6 @@
  * along with this program.  If not, see <http://www.gnu.org/licenses/>.
  */
 #include <bitcoin/system/unicode/utf8_everywhere/stdio.hpp>
-
```
evoskuil, line 20: Style: blank line after primary own header include.

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -80,47 +78,65 @@ inline void set_binary_stdio(FILE* file) THROWS
         throw runtime_exception{ "Could not set STDIO to binary mode." };
 }
 
-void set_console_echo() NOEXCEPT
+#ifndef _WIN32
```
evoskuil, line 81: Use `HAVE_MSC` as use of vendor PP defines is unsafe, for example:
```
#if defined(_MSC_VER) && !defined(HAVE_CLANG)
    #define HAVE_MSC
#endif
```
See `have.hpp`.

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -80,47 +78,65 @@ inline void set_binary_stdio(FILE* file) THROWS
...
-    const auto handle = GetStdHandle(STD_INPUT_HANDLE);
-    DWORD mode{};
-    GetConsoleMode(handle, &mode);
-    SetConsoleMode(handle, mode | ENABLE_ECHO_INPUT);
-}
+    termios terminal{};
+    tcgetattr(0, &terminal);
+    terminal.c_lflag &= ~ECHO;
+    tcsetattr(STDIN_FILENO, TCSAFLUSH, &terminal);
+}}
```
evoskuil, line 88: Syntax error ( extraneous `}` )..

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -80,47 +78,65 @@ inline void set_binary_stdio(FILE* file) THROWS
...
 
-void unset_console_echo() NOEXCEPT
+void unset_console_echo()
```
evoskuil, line 90: This function is now defined twice.

src/unicode/utf8_everywhere/stdio.cpp
```diff
@@ -80,47 +78,65 @@ inline void set_binary_stdio(FILE* file) THROWS
...
-    const auto handle = GetStdHandle(STD_INPUT_HANDLE);
-    DWORD mode{};
-    GetConsoleMode(handle, &mode);
-    SetConsoleMode(handle, mode & ~ENABLE_ECHO_INPUT);
+    termios terminal{};
+    tcgetattr(0, &terminal);
+    terminal.c_lflag &= ~ECHO;
+    tcsetattr(STDIN_FILENO, TCSAFLUSH, &terminal);
 }
+#else
```
evoskuil, line 97: Echo for Windows is hereby entirely disabled.

evoskuil (commented):
I'll push changes to your repo.

PR #1829, opened 2026-05-08
fix(block::check) fix dead code on verification of empty block

KY-U:
The `is_empty()` verification inside the `check()` function can never be `true`, since it is nested inside the `if (is_first_non_coinbase())` condition, that fails for empty blocks, so `check()` never actually signaled on an empty block.


evoskuil:
That’s more than dead code, thanks!

This is the offending commit.

https://github.com/libbitcoin/libbitcoin-system/commit/d0c10c01c126e8fdd341d1cdf0014ac947432c41

evoskuil:
Thanks also for adding the test. Completing full unit test  coverage for chain and machine objects is part of the release checklist.

PR #1836, opened 2026-05-20
CONSENSUS: fix regression and fault in taproot integration.

evoskuil:
The taproot implementation opcodes still requires unit testing. There is a fix to an implementation error and a regression in multisig related to the taproot integration.

### DIFF
--- a/include/bitcoin/system/error/op_error_t.hpp
+++ b/include/bitcoin/system/error/op_error_t.hpp
@@ -97,17 +97,19 @@ enum op_error_t : uint8_t
     op_sha256,
     op_hash160,
     op_hash256,
+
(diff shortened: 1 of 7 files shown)


thecodefactory:
LGTM.  With this and a proper config, testnet3 syncs w/o checkpoints.

evoskuil:
Mainnet block 0-950,000 validated here as well.

evoskuil:
The `op_check_multisig_verify()` regression patched in this PR was independently discovered and reported less than 24 hours after the creation of this PR, by @KY-U using [bitcoinfuzz](https://github.com/bitcoinfuzz/bitcoinfuzz).

PR #1840, opened 2026-05-22
fix(interpreter): fail op_check_sig when less than 2 items on stack

KY-U:
OP_CHECK_SIG called with less than 2 items on stack should fail script execution.


evoskuil:
Looks good, thanks again!

PR #1850, opened 2026-05-27
Only immediately fail op_check_sig_verify execution with empty key when BIP342 is activted.

KY-U:
Empty key is only a hard fail under BIP342. Legacy OPCHECKSIGVERIFY does eventually fail, so this change may be somewhat pedantic, but it matches more closely with Core's behaviour.


evoskuil (changes requested):
Looks good, but the added tests cases are failing.

KY-U:
Fixed the error. I must have recompiled just the test executable without recompiling the libbitcoin-system lib and thought it was passing. Sorry about that, skill issue.

PR #1855, opened 2026-05-30
crypto: GPU/CPU batch signature verification (UltrafastSecp256k1 backend)

shrec:
## Summary

Adds **batch signature verification** to `bitcoin/system/crypto/secp256k1` —
`ecdsa::batch_verify` and `schnorr::batch_verify` — targeting IBD / historical
block validation, where signature checks are ~95% of script-validation cost.
The node packs the `sig / key / sighash` triples extracted from script execution
(`CHECKSIG`, `CHECKMULTISIG`, Taproot `CHECKSIG` / `CHECKSIGADD`) into one byte
buffer and gets a per-row pass/fail array back — one call per kind. This is the
batch entry point discussed for the optional **UltrafastSecp256k1** secp256k1
backend (GPU/CPU accelerated, with a mandatory CPU fallback). Not a mempool path.

## Packed row layout

Homogeneous and uniform-size; ECDSA and Schnorr stack independently.

| Kind | Record | Bytes |
|------|--------|-------|
| ECDSA | `32 hash \| 33 point \| 64 signature` | 129 |
| Schnorr | `32 x-only key \| 32 hash \| 64 signature` | 128 |

Each row may carry an optional trailing **opaque key** of variable `key_size`
bytes (e.g. a 3-byte block id or 4-byte tx id). It is never interpreted — carried
only so an invalid row maps back to its block/tx without a second side table.
`key_size == 0` disables the column (stride == record size).

```cpp
bool ecdsa::batch_verify  (const data_slice& rows, size_t count, size_t key_size,
                           std_vector<uint8_t>& results) NOEXCEPT;
bool schnorr::batch_verify(const data_slice& rows, size_t count, size_t key_size,
                           std_vector<uint8_t>& results) NOEXCEPT;
```
`results` is sized to `count`; `results[i]` is non-zero iff row `i` verifies. A
structurally-invalid row reports `0` (not an error); `false` is returned only on
malformed input (rows too short).

## Two backends behind one signature

- **`WITH_ULTRAFAST`** — dispatch to the UltrafastSecp256k1 batch bridge
  (`ufsecp_lbtc_verify_ecdsa` / `_verify_schnorr`): GPU when a backend is linked,
  with a mandatory engine-side CPU fallback. One controller per worker thread.
- **otherwise** — a parallel `std::for_each` (the vendored poolSTL `par` policy)
  over the singular `verify_signature`. This is the **consensus reference** the
  GPU path must match bit-for-bit, and a benchmark baseline that also benefits
  the non-ultrafast build.

## Build

`with-ultrafast` (CMake) routes the secp256k1 backend to the engine's drop-in
libsecp256k1 ABI shim — single-signature ops compile and link **unchanged** — and
defines `WITH_ULTRAFAST` so the batch wrappers use the bridge. Engine at
`deps/UltrafastSecp256k1` (submodule) or `-DUltrafastSecp256k1_DIR=<path>`.

```bash
cmake -S builds/cmake -B build -Dwith-ultrafast=ON -DCMAKE_BUILD_TYPE=Release
```

Default build (`with-ultrafast=OFF`) is unchanged: stock libsecp256k1 +
the parallel CPU `batch_verify`.

## Tests

`test/crypto/elliptic_curve.cpp`: ECDSA and Schnorr `batch_verify` — all-valid,
one-invalid, and an opaque-key-column case. These exercise the CPU reference path.

## Notes

This is a **proposal / proof-of-concept** for review (per our discussion): the
design is the one we converged on — one unified table per call, separate
ECDSA/Schnorr, variable opaque key column, mandatory CPU fallback. The GPU-vs-CPU
**consensus differential** gate (GPU == CPU == libsecp256k1, bit-for-bit) lives
on the engine side; the IBD marshalling is the node side. A BIP-352 silent-payment
scan entry point (`ufsecp_lbtc_sp_scan`) is also exposed by the bridge for the
Electrum/server path. See `doc/ultrafast-secp256k1-backend.md`.

### DIFF
--- a/builds/cmake/CMakeLists.txt
+++ b/builds/cmake/CMakeLists.txt
@@ -28,6 +28,7 @@ option( enable-sse41 "Use SSE4.1 hardware instructions." OFF )
 option( enable-shani "Use Intel/ARM SHA Extensions." OFF )
 option( with-tests "Build tests." ON )
 option( with-examples "Build examples." ON )
+option( with-ultrafast "Use UltrafastSecp256k1 (GPU/CPU batch signature verification) as the secp256k1 backend." OFF )
 
 #------------------------------------------------------------------------------
 # Dependencies.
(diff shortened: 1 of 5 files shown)


evoskuil:
Retain for reference.

PR #1856, opened 2026-06-01
add silent payments primitives for an index and electrum endpoint

josibake:
This is the first of a few PRs for adding a silent payments tweak index and electrum scanning endpoint to libbitcoin-server. The branches are ready, but I'm polishing the code a bit more before publishing. Publishing this one as a start to get some early feedback, as changes here will affect the downstream PRs substantially.

I've added a silent payments "module," but I would consider this proof of concept for right now. I think it would make more sense long term to use the implementation in Ultrafastsecp for its GPU support, if it is deemed safe and a good fit for this project. Alternatively, this could be swapped out in favour of the silent payments libsecp module once it merges.

I'd direct reviewers instead to look at the libbitcoin specific bits:

* There is a small bug fix and some tests for a taproot parsing error (uncovered when I added the silent payment test vectors)
* There is a more generic bech32 parser that allows the caller to specify which checksum they want to use , instead of relying on the witness version
* There is an ec_add helper for uncompressed keys (relevant for silent payments but perhaps generally useful if there are other places where keys are repeatedly decompressed)

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -147,6 +147,7 @@ src_libbitcoin_system_la_SOURCES = \
     src/wallet/addresses/bitcoin_uri.cpp \
     src/wallet/addresses/payment_address.cpp \
     src/wallet/addresses/qr_code.cpp \
+    src/wallet/addresses/silent_payment.cpp \
     src/wallet/addresses/stealth_address.cpp \
     src/wallet/addresses/stealth_receiver.cpp \
     src/wallet/addresses/stealth_sender.cpp \
@@ -438,6 +439,8 @@ test_libbitcoin_system_test_SOURCES = \
     test/wallet/addresses/checked.cpp \
     test/wallet/addresses/payment_address.cpp \
     test/wallet/addresses/qr_code.cpp \
+    test/wallet/addresses/silent_payment.cpp \
+    test/wallet/addresses/silent_payment_vectors.hpp \
     test/wallet/addresses/stealth_address.cpp \
     test/wallet/addresses/stealth_receiver.cpp \
     test/wallet/addresses/stealth_sender.cpp \
@@ -983,6 +986,7 @@ include_bitcoin_system_wallet_addresses_HEADERS = \
     include/bitcoin/system/wallet/addresses/checked.hpp \
     include/bitcoin/system/wallet/addresses/payment_address.hpp \
     include/bitcoin/system/wallet/addresses/qr_code.hpp \
+    include/bitcoin/system/wallet/addresses/silent_payment.hpp \
     include/bitcoin/system/wallet/addresses/stealth_address.hpp \
     include/bitcoin/system/wallet/addresses/stealth_receiver.hpp \
     include/bitcoin/system/wallet/addresses/stealth_sender.hpp \
@@ -1035,4 +1039,3 @@ target_examples = \
     examples/libbitcoin-system-examples
 
 examples: ${target_examples}
-
--- a/include/bitcoin/system/impl/chain/script_patterns.ipp
+++ b/include/bitcoin/system/impl/chain/script_patterns.ipp
@@ -181,7 +181,7 @@ constexpr bool script::is_pay_witness_taproot_pattern(
     const operations& ops) NOEXCEPT
 {
     return ops.size() == 2
-        && ops[0].code() == opcode::push_size_1
+        && ops[0].code() == opcode::push_positive_1
         && ops[1].code() == opcode::push_size_32;
 }
 
@@ -364,7 +364,7 @@ inline operations script::to_pay_witness_taproot_pattern(
 {
     return
     {
-        { opcode::push_size_1 },
+        { opcode::push_positive_1 },
         { to_chunk(hash), false }
     };
 }
--- a/include/bitcoin/system/impl/chain/script_patterns.ipp
+++ b/include/bitcoin/system/impl/chain/script_patterns.ipp
@@ -216,6 +216,15 @@ constexpr bool script::is_sign_key_hash_pattern(const operations& ops) NOEXCEPT
         && is_public_key(ops[1].data());
 }
 
+constexpr bool script::is_sign_witness_key_hash_pattern(
+    const operations& ops) NOEXCEPT
+{
+    return ops.size() == 1
+        && ops[0].code() == opcode::push_size_22
+        && ops[0].data().size() == short_hash_size + 2u
+        && ops[0].data()[0] == 0x00
+        && ops[0].data()[1] == short_hash_size;
+}
 
 // Ambiguous with is_sign_key_hash when second/last op is a public key.
 // Ambiguous with is_sign_public_key_pattern when only op is endorsement.
--- a/include/bitcoin/system/wallet/addresses/silent_payment.hpp
+++ b/include/bitcoin/system/wallet/addresses/silent_payment.hpp
@@ -1,0 +1,114 @@
+/**
+ * Copyright (c) 2011-2026 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#ifndef LIBBITCOIN_SYSTEM_WALLET_ADDRESSES_SILENT_PAYMENT_HPP
+#define LIBBITCOIN_SYSTEM_WALLET_ADDRESSES_SILENT_PAYMENT_HPP
+
+#include <optional>
+#include <bitcoin/system/crypto/secp256k1.hpp>
+#include <bitcoin/system/define.hpp>
+
+namespace libbitcoin {
+namespace system {
+namespace chain {
+
+class transaction;
+
+} // namespace chain
+namespace wallet {
+namespace silent_payment {
+
+struct BC_API pay_witness_taproot_output
+{
+    uint32_t index{};
+    ec_xonly key{};
+};
+
+struct BC_API scan_record
+{
+    ec_compressed prevouts_summary{};
+    std_vector<pay_witness_taproot_output> outputs{};
+};
+
+struct BC_API scan_match
+{
+    uint32_t index{};
+    ec_secret tweak{};
+    std::optional<uint32_t> label{};
+
+    bool operator<(const scan_match& other) const NOEXCEPT;
+};
+
+typedef std_vector<scan_match> scan_matches;
+
+class BC_API scanner
+{
+public:
+    DEFAULT_COPY_MOVE_DESTRUCT(scanner);
+
+    scanner(const ec_secret& scan_secret, const ec_compressed& spend_public,
+        const std_vector<uint32_t>& labels) NOEXCEPT;
+
+    operator bool() const NOEXCEPT;
+
+    /// Scan outputs from one transaction, with unique output indexes.
+    bool scan(scan_matches& out, const ec_compressed& prevouts_summary,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+    bool scan(scan_matches& out, const ec_uncompressed& prevouts_summary,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+
+    /// Match outputs from one transaction, with unique output indexes.
+    bool match(bool& out, const ec_compressed& prevouts_summary,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+    bool match(bool& out, const ec_uncompressed& prevouts_summary,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+
+private:
+    bool valid() const NOEXCEPT;
+    bool to_shared_secret(ec_compressed& out,
+        const ec_compressed& prevouts_summary) const NOEXCEPT;
+    bool to_shared_secret(ec_compressed& out,
+        const ec_uncompressed& prevouts_summary) const NOEXCEPT;
+    bool scan_shared(scan_matches& out, const ec_compressed& shared_secret,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+    bool match_shared(bool& out, const ec_compressed& shared_secret,
+        const std_vector<pay_witness_taproot_output>& outputs)
+        const NOEXCEPT;
+
+    ec_secret scan_secret_{};
+    ec_uncompressed spend_point_{};
+    std_vector<uint32_t> labels_{};
+    secret_list label_tweaks_{};
+    uncompressed_list label_points_uncompressed_{};
+    bool valid_{};
+};
+
+BC_API bool compute_scan_record(scan_record& out,
+    const chain::transaction& tx) NOEXCEPT;
+
+} // namespace silent_payment
+} // namespace wallet
+} // namespace system
+} // namespace libbitcoin
+
+#endif
--- a/src/crypto/secp256k1.cpp
+++ b/src/crypto/secp256k1.cpp
@@ -211,6 +211,26 @@ bool ec_add(ec_compressed& left, const ec_uncompressed& right) NOEXCEPT
     return compress(out, right) && ec_add(left, out);
 }
 
+bool ec_add(ec_uncompressed& left, const ec_uncompressed& right) NOEXCEPT
+{
+    const auto context = ec_context_verify::context();
+
+    secp256k1_pubkey left_key;
+    secp256k1_pubkey right_key;
+    if (!parse(context, left_key, left) || !parse(context, right_key, right))
+        return false;
+
+    const secp256k1_pubkey* keys[]
+    {
+        &left_key,
+        &right_key
+    };
+
+    secp256k1_pubkey out;
+    return secp256k1_ec_pubkey_combine(context, &out, keys, 2) ==
+        ec_success && serialize(context, left, out);
+}
+
 // parse, combine, serialize
 bool ec_sum(ec_compressed& out, const compressed_list& points) NOEXCEPT
 {
--- a/src/wallet/addresses/silent_payment.cpp
+++ b/src/wallet/addresses/silent_payment.cpp
@@ -1,0 +1,588 @@
+/**
+ * Copyright (c) 2011-2026 libbitcoin developers (see AUTHORS)
+ *
+ * This file is part of libbitcoin.
+ *
+ * This program is free software: you can redistribute it and/or modify
+ * it under the terms of the GNU Affero General Public License as published by
+ * the Free Software Foundation, either version 3 of the License, or
+ * (at your option) any later version.
+ *
+ * This program is distributed in the hope that it will be useful,
+ * but WITHOUT ANY WARRANTY; without even the implied warranty of
+ * MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+ * GNU Affero General Public License for more details.
+ *
+ * You should have received a copy of the GNU Affero General Public License
+ * along with this program.  If not, see <http://www.gnu.org/licenses/>.
+ */
+#include <bitcoin/system/wallet/addresses/silent_payment.hpp>
+
+#include <algorithm>
+#include <iterator>
+#include <utility>
+#include <bitcoin/system/chain/enums/script_pattern.hpp>
+#include <bitcoin/system/chain/enums/script_version.hpp>
+#include <bitcoin/system/chain/input.hpp>
+#include <bitcoin/system/chain/operation.hpp>
+#include <bitcoin/system/chain/output.hpp>
+#include <bitcoin/system/chain/point.hpp>
+#include <bitcoin/system/chain/script.hpp>
+#include <bitcoin/system/chain/tapscript.hpp>
+#include <bitcoin/system/chain/transaction.hpp>
+#include <bitcoin/system/crypto/secp256k1.hpp>
+#include <bitcoin/system/data/data.hpp>
+#include <bitcoin/system/define.hpp>
+#include <bitcoin/system/hash/hash.hpp>
+#include <bitcoin/system/radix/base_16.hpp>
+#include <bitcoin/system/stream/stream.hpp>
+
+namespace libbitcoin {
+namespace system {
+namespace wallet {
+namespace silent_payment {
+
+BC_PUSH_WARNING(NO_THROW_IN_NOEXCEPT)
+
+using chain::input;
+using chain::point;
+using chain::script;
+using chain::script_pattern;
+using chain::script_version;
+using chain::tapscript;
+using chain::transaction;
+
+static constexpr auto bip341_nums_key = base16_array(
+    "50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0");
+static constexpr auto ec_odd_sign = add1(ec_even_sign);
+static constexpr size_t maximum_outputs = 2323;
+
+typedef data_array<point::serialized_size()> outpoint_data;
+
+static outpoint_data to_outpoint_data(const point& value) NOEXCEPT
+{
+    outpoint_data out{};
+    stream::out::fast stream{ out };
+    write::bytes::fast writer{ stream };
+    value.to_data(writer);
+    return out;
+}
+
+static bool outpoint_data_less(const outpoint_data& left,
+    const outpoint_data& right) NOEXCEPT
+{
+    return std::lexicographical_compare(left.begin(), left.end(),
+        right.begin(), right.end());
+}
+
+static bool is_compressed_key_sign(uint8_t value) NOEXCEPT
+{
+    return value == ec_even_sign || value == ec_odd_sign;
+}
+
+static bool copy_xonly(ec_xonly& out, const data_chunk& value) NOEXCEPT
+{
+    if (value.size() != ec_xonly_size)
+        return false;
+
+    std::copy_n(value.begin(), ec_xonly_size, out.begin());
+    return true;
+}
+
+static bool copy_compressed(ec_compressed& out,
+    const data_chunk& value) NOEXCEPT
+{
+    if (!is_compressed_key(value))
+        return false;
+
+    std::copy_n(value.begin(), ec_compressed_size, out.begin());
+    return true;
+}
+
+static ec_xonly to_xonly(const ec_uncompressed& point) NOEXCEPT
+{
+    ec_xonly out{};
+    std::copy_n(std::next(point.begin()), ec_xonly_size, out.begin());
+    return out;
+}
+
+static bool shared_secret_tweak(ec_secret& out,
+    const ec_compressed& shared_secret, uint32_t index) NOEXCEPT
+{
+    stream::out::fast stream{ out };
+    hash::sha256t::fast<"BIP0352/SharedSecret"> sink{ stream };
+    sink.write_bytes(shared_secret);
+    sink.write_4_bytes_big_endian(index);
+    sink.flush();
+    return verify(out);
+}
+
+static bool contains_match(const scan_matches& matches, uint32_t index) NOEXCEPT
+{
+    for (const auto& match: matches)
+        if (match.index == index)
+            return true;
+
+    return false;
+}
+
+static bool find_unmatched_output(uint32_t& out_index,
+    const std_vector<pay_witness_taproot_output>& outputs, const ec_xonly& key,
+    const scan_matches& matches) NOEXCEPT
+{
+    for (const auto& output: outputs)
+    {
+        if (output.key != key || contains_match(matches, output.index))
+            continue;
+
+        out_index = output.index;
+        return true;
+    }
+
+    return false;
+}
+
+static bool contains_output(
+    const std_vector<pay_witness_taproot_output>& outputs,
+    const ec_xonly& key) NOEXCEPT
+{
+    for (const auto& output: outputs)
+        if (output.key == key)
+            return true;
+
+    return false;
+}
+
+static bool extract_last_witness_key(ec_compressed& out,
+    const input& in) NOEXCEPT
+{
+    const auto& stack = in.witness().stack();
+    if (stack.empty() || !stack.back())
+        return false;
+
+    return copy_compressed(out, *stack.back());
+}
+
+static bool extract_p2pkh_key(ec_compressed& out, const input& in) NOEXCEPT
+{
+    const auto& ops = in.prevout->script().ops();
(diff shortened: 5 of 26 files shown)


test/hash/checksum.cpp
```diff
@@ -247,6 +247,28 @@ BOOST_AUTO_TEST_CASE(checksum__bech32_build_checked__five_program_bytes__expecte
     BOOST_REQUIRE_EQUAL(encode_base32(checked), "qqypqxpq939vyak");
 }
 
+BOOST_AUTO_TEST_CASE(checksum__bech32_build_checked__version_zero_bech32m__expected)
+{
+    const auto checked = bech32_build_checked(0, {}, "abcdef",
+        checksum_constant::bech32m);
```
evoskuil: Test style: avoid wrapping lines in TEST_CASEs. Leave long or reduce using intermediate vars. Multiline statements are generally limited to structure defines.

include/bitcoin/system/chain/silent_payment.hpp
evoskuil, line 1: Move from chain to wallet/addresses (see stealth utils for similar pattern).

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+        if (output.key == key)
+            return true;
+
+    return false;
+}
+
+static void remove_output(std_vector<output>& outputs, uint32_t index) NOEXCEPT
+{
+    std::erase_if(outputs,
+        [index](const auto& output) NOEXCEPT
```
evoskuil: Style: keep lambda on the top line and shift {} fully left.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+        {
+            return output.index == index;
+        });
+}
+
+static bool is_p2sh_p2wpkh_redeem(const script& sig) NOEXCEPT
```
evoskuil: Move to chain::script pattern.

evoskuil: Avoid referring to script as "sig" or "pubkey". Use input script or output script. But in this case the test should be over the op stack, since requiring the script is too restrictive.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+{
+    const auto& ops = sig.ops();
+    return ops.size() == one
+        && ops.front().is_push()
+        && ops.front().data().size() == 22u
+        && ops.front().data().at(0) == 0x00
+        && ops.front().data().at(1) == 0x14;
+}
+
+static bool extract_last_witness_key(ec_compressed& out,
```
evoskuil: Is this the annex? If so this is `witness.annex()`.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+
+    const auto minimum = std::min_element(inputs.begin(), inputs.end(),
+        input_outpoint_less);
+    if (minimum == inputs.end())
+        return false;
+
+    const auto outpoint = to_outpoint_data((*minimum)->point());
+
+    hash_digest digest{};
+    stream::out::fast stream{ digest };
```
evoskuil: Construct on out.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    hash::sha256t::fast<"BIP0352/Inputs"> sink{ stream };
+    sink.write_bytes(outpoint);
```
evoskuil: use `point.to_data(writer)`.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    sink.write_bytes(sum);
+    sink.flush();
+
+    out = digest;
+    return verify(out);
+}
+
+bool compute_scan_record(record& out, const transaction& tx) NOEXCEPT
+{
+    out = {};
```
evoskuil: In general, no need to initialize returns when there is an independent failure result (return bool). If the caller depends on the result in case of fail it's his bug. truth in one place.

If there's a case where the return matters in the case of a false result that should be clearly commented.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    if (!ec_sum(sum, keys) || !compute_input_hash(input_hash, tx, sum))
+        return false;
+
+    out.tweak_key = sum;
+    return ec_multiply(out.tweak_key, input_hash);
+}
+
+bool compute_scan_records(records& out, const block& block) NOEXCEPT
+{
+    out.clear();
```
evoskuil: same

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+                matched = true;
+                break;
+            }
+        }
+
+        if (!matched)
+            break;
+    }
+
+    std::sort(out.begin(), out.end(),
```
evoskuil: Style: unwrap line

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+        [](const auto& left, const auto& right) NOEXCEPT
+        {
+            return left.index < right.index;
```
evoskuil: overload < operator to avoid explicitness here.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+{
+    stream::out::fast stream{ out };
+    hash::sha256t::fast<"BIP0352/SharedSecret"> sink{ stream };
+    sink.write_bytes(shared_secret);
+    sink.write_4_bytes_big_endian(index);
+    sink.flush();
+    return verify(out);
+}
+
+static bool find_output(uint32_t& out_index, const std_vector<output>& outputs,
```
evoskuil: `std_vector<output>` is `chain::outputs`.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    const ec_compressed& spend_public, const std_vector<uint32_t>& labels,
+    const ec_compressed& tweak_key, const std_vector<output>& outputs) NOEXCEPT
+{
+    out.clear();
+
+    scan_context context{};
+    if (!create_scan_context(context, scan_secret, spend_public, labels))
+        return false;
+
+    return scan_outputs(out, context, tweak_key, outputs);
```
evoskuil: Combine into one return.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    out.clear();
+
+    if (!valid_scan_context(context))
+        return false;
+
+    ec_compressed shared_secret{};
+    if (!to_shared_secret(shared_secret, context, tweak_key))
+        return false;
+
+    return scan_outputs_shared(out, context, shared_secret, outputs);
```
evoskuil: Combine 2 into one return. If failure is the cold/rare path then could combine all three.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
```
evoskuil: Combine into one return.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    out = false;
+
+    if (!valid_scan_context(context))
+        return false;
+
+    ec_compressed shared_secret{};
+    if (!to_shared_secret(shared_secret, context, tweak_key))
+        return false;
+
+    return match_outputs_shared(out, context, shared_secret, outputs);
```
evoskuil: Combine into one return.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
```
evoskuil: Combine into one return.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+{
+    const auto& ops = in.prevout->script().ops();
+    if (!script::is_pay_key_hash_pattern(ops))
+        return false;
+
+    const auto bytes = in.script().to_data(false);
+    if (bytes.size() < ec_compressed_size)
+        return false;
+
+    // P2PKH scriptSigs are malleable, so scan backward for the compressed key
```
evoskuil: scriptSigs :<

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    for (auto offset = bytes.size() - ec_compressed_size + one; offset > zero;
+        --offset)
+    {
+        const auto start = std::next(bytes.begin(), offset - one);
+        const data_chunk key{ start, std::next(start, ec_compressed_size) };
+        if (!copy_compressed(out, key))
+            continue;
+
+        const auto digest = bitcoin_short_hash(key);
+        if (std::equal(digest.begin(), digest.end(), ops.at(2).data().begin()))
```
evoskuil: use simple compare against array_cast<>&

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    return false;
+}
+
+static bool extract_taproot_key(ec_compressed& out, const input& in) NOEXCEPT
+{
+    if (is_nums_internal_key(in))
+        return false;
+
+    const auto& program = in.prevout->script().witness_program();
+    if (is_null(program) || program->size() != ec_xonly_size)
```
evoskuil: We probably aren't consistent, but can use !ptr for shared ptrs (since they have a bool operator).

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+        return false;
+
+    ec_xonly xonly{};
+    if (!copy_xonly(xonly, *program))
+        return false;
+
+    return to_even_compressed(out, xonly);
```
evoskuil: Combined returns.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+}
+
+static bool extract_taproot_output(output& out, const chain::output& point,
+    uint32_t index) NOEXCEPT
+{
+    if (point.script().output_pattern() != script_pattern::pay_witness_v1_taproot)
+        return false;
+
+    const auto& program = point.script().witness_program();
+    if (is_null(program) || !copy_xonly(out.key, *program))
```
evoskuil: same

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+        return false;
+
+    ec_xonly xonly{};
+    if (!copy_xonly(xonly, *program))
+        return false;
+
+    return to_even_compressed(out, xonly);
+}
+
+static bool extract_taproot_output(output& out, const chain::output& point,
```
evoskuil: Misnomer: param name conflates output with point type. So for example `point.script()` below is jarring. Is one the output and the other the previous output (prevout)?

Inconsistent namespace `output&` vs. `chain::output&`.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    uint32_t index) NOEXCEPT
+{
+    if (point.script().output_pattern() != script_pattern::pay_witness_v1_taproot)
```
evoskuil: If only looking for one pattern it's more efficient to just test that one vs. testing them all.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+    const auto& stack = in.witness().stack();
+
+    // BIP352 only skips taproot script-path spends whose control block internal
+    // key is the BIP341 NUMS point H.
+    if (stack.size() <= one)
+        return false;
+
+    auto control = stack.back();
+    if (annex::is_annex_pattern(stack))
+    {
```
evoskuil: Style: remove {}.

src/chain/silent_payment.cpp
```diff
@@ -0,0 +1,639 @@
...
+}
+
+bool compute_input_hash(ec_secret& out, const transaction& tx,
+    const ec_compressed& sum) NOEXCEPT
+{
+    const auto& inputs = *tx.inputs_ptr();
+    if (inputs.empty())
+        return false;
+
+    const auto minimum = std::min_element(inputs.begin(), inputs.end(),
```
evoskuil: Simplify using < operator overload.

Makefile.am
```diff
@@ -51,6 +51,7 @@ src_libbitcoin_system_la_SOURCES = \
     src/chain/point.cpp \
     src/chain/script.cpp \
     src/chain/script_extract.cpp \
+    src/chain/silent_payment.cpp \
```
evoskuil: Add/removing/moving sources requires updates to all builds. Maintainers do this using the libbitcoin-build generator. E.g.

`libbitcoin-build\generate4.sh system`
`libbitcoin-build\generate4.cmd system`

> This executes the iMatix GSL code generator.
> See https://github.com/imatix/gsl for details.

include/bitcoin/system/chain/silent_payment.hpp
evoskuil, line 1: This is very well done, however the large number of free functions that appear highly interrelated leaves me wondering if this could be quite a bit simpler from the calling and namespace perspective as a stateful class.

test/wallet/addresses/silent_payment.cpp
evoskuil, line 1: There are some style issues in both SP test sources, though minor.

I'd note that there are no unit tests for the individual free functions that constitute the API. The test strategy is what we call component (large unit aggregates) test as opposed to unit test. This is sometimes necessary, or easier, but we try to limit this to validation against public vectors (as it appears to be here) but also test each unit individually.

This allows failure/regression detection to identify the location of the failure, and allows the unit tests to clearly express the intent of each public (and protected) unit.  Also the test setup for component (and functional - e2e) tests tends to be at least as complex as the code being tested, which makes it impossible to verify that the tests are actually correct (this is why unit tests are required to be fully declarative).

So I'd suggest in addition to the verification against public vectors below, a set of unit tests for the methods of the public interface.

evoskuil (changes requested):
Excellent overall.

include/bitcoin/system/chain/silent_payment.hpp
josibake: Thanks for pointing me to the stealth implementation. I decided to pull heavily from that design while rewriting the silent payments API as a class and I am much happier with the result.

src/chain/silent_payment.cpp
josibake: Nice, I like this a lot better.

Makefile.am
josibake: I think I did this correctly, but haven't spent enough time fully understanding the build system.

include/bitcoin/system/chain/silent_payment.hpp
josibake: Yeah, at first pass I think I was focusing too much on translating code I had written previously to the libbitcoin idioms, more as a "correctness" check. But after using this PR in the database and server, and looking at the stealth example, I started to realise it was overly complicated and decided to just follow the stealth pattern. Looks a lot cleaner now, I think. I guess one open question in my mind is: should the silent payments API here be a general "everything you need for silent payments, from address generation to sending to receiving," but I don't think thats a question for this PR. I think for now keeping it scoped to "everything you need to implement scanning" is sufficient.

test/wallet/addresses/silent_payment.cpp
josibake: Spent some more time looking at your unit tests. Coupled with the smaller surface area of the API, I think this is cleaner. I'm still pulling the raw data from the vectors, for both the component tests and the unit tests, but I think thats fine for now.

src/chain/silent_payment.cpp
josibake: I don't think thats appropriate here, but I tried to make it more explicit with an explanation and `std::lexicographical_compare`. This is correct and, imo, easier to verify visually that the implementation matches the BIP (which specifies finding the lexicographically lowest outpoint based on its byte serialisation).

josibake:
I believe I've addressed all of the review comments. I also rewrote this to be more purposeful (vs a generic silent payments toolkit), which is the correct approach at this stage, imo. I also tried to apply the review comments more generally while writing the new code, but still might have missed things.

josibake:
I found myself jumping back to the code I wrote to remind myself what names meant, which indicates super smelly code. Specifically, `scan_output`, which is meant to be a taproot output, could have meant an output to be scanned, an function to scan an output, or an output that was emitted by a scan (a found output). I tried to match the existing nomenclature with `pay_witness_taproot_output`. Considering its a bit long, I'd be open to other suggestions, e.g. `pay_witness_v1_output`, `taproot_output`, etc.

I also changed `tweak_key` to the libsecp256k1 module nomenclature `prevouts_summary`. I think this name is much better and clearly describes what the data is: the lexicographically smallest outpoint multiplied by with the sum of the prevout public keys. `tweak_key` doesn't tell where the data came from or what it is, and also could be confused with taproot tweak keys, BIP32 key tweaking, etc.

src/chain/silent_payment.cpp
evoskuil: The initialize for failures is part of the "defensive programming" practice of hiding bugs.

include/bitcoin/system/impl/chain/script_patterns.ipp
```diff
@@ -216,6 +216,15 @@ constexpr bool script::is_sign_key_hash_pattern(const operations& ops) NOEXCEPT
         && is_public_key(ops[1].data());
 }
 
+constexpr bool script::is_sign_witness_key_hash_pattern(
+    const operations& ops) NOEXCEPT
+{
+    return ops.size() == 1
+        && ops[0].code() == opcode::push_size_22
+        && ops[0].data().size() == short_hash_size + 2u
+        && ops[0].data()[0] == 0x00
```
evoskuil, line 225: Is this value (0x00) in this usage a symbol defined anywhere in system?

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+#include <optional>
+#include <bitcoin/system/crypto/secp256k1.hpp>
+#include <bitcoin/system/define.hpp>
+
+namespace libbitcoin {
+namespace system {
+namespace chain {
+
+class transaction;
```
evoskuil, line 30: Just use `class chain::transaction` inside of wallet namespace.

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+} // namespace chain
+namespace wallet {
+namespace silent_payment {
```
evoskuil, line 34: Drop the namespace and use the class as the namespace for these.

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+    uint32_t index{};
+    ec_secret tweak{};
+    std::optional<uint32_t> label{};
+
+    bool operator<(const scan_match& other) const NOEXCEPT;
+};
+
+typedef std_vector<scan_match> scan_matches;
+
+class BC_API scanner
```
evoskuil, line 59: Class should match file name. Also as a rule we try to limit to one class or enum per source (some exceptions of course). Suggest naming the class `silent_payment`.

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+{
+    ec_compressed prevouts_summary{};
+    std_vector<pay_witness_taproot_output> outputs{};
+};
+
+struct BC_API scan_match
+{
+    uint32_t index{};
+    ec_secret tweak{};
+    std::optional<uint32_t> label{};
```
evoskuil, line 52: Since this is an integral you can prob get away with a sentinel, which reduces the size of the struct, but really not significant.

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+class BC_API scanner
+{
+public:
+    DEFAULT_COPY_MOVE_DESTRUCT(scanner);
+
+    scanner(const ec_secret& scan_secret, const ec_compressed& spend_public,
+        const std_vector<uint32_t>& labels) NOEXCEPT;
+
+    operator bool() const NOEXCEPT;
```
evoskuil, line 67: This is correct and one of our inconsistencies. Chain types define a method for this while wallet types define the operator.

include/bitcoin/system/wallet/addresses/silent_payment.hpp
```diff
@@ -0,0 +1,114 @@
...
+namespace system {
+namespace chain {
+
+class transaction;
+
+} // namespace chain
+namespace wallet {
+namespace silent_payment {
+
+struct BC_API pay_witness_taproot_output
```
evoskuil, line 36: If you swap key and index this is the `point` class (and `points`). operator and ref types are all defined, as are de/serialization. Since these are ec keys and not hashes that use would overload the name, but could use an alias to define this (and the first part of `scan_match` below). This does add the `value_` member, but if you don't reference it, it it will compile away.

You're right about the name, it's a bit long given its repeating usage and limited context. You could use `scan_output` as these represents outputs and are the elements of a `scan_record`, and alias `using `scan_records = std_vector<scan_record>`.

src/crypto/secp256k1.cpp
```diff
@@ -211,6 +211,26 @@ bool ec_add(ec_compressed& left, const ec_uncompressed& right) NOEXCEPT
...
+    secp256k1_pubkey left_key;
+    secp256k1_pubkey right_key;
+    if (!parse(context, left_key, left) || !parse(context, right_key, right))
+        return false;
+
+    const secp256k1_pubkey* keys[]
+    {
+        &left_key,
+        &right_key
+    };
```
evoskuil, line 227: Could be: `const secp256k1_pubkey* keys[]{ &left_key, &right_key };`

But we generally try to avoid C-style arrays, suggest:

```cpp
    std::array<secp256k1_pubkey, 2> keys{};
    if (!parse(context, keys.front(), left) ||
        !parse(context, keys.back(), right))
        return false;

    secp256k1_pubkey out;
    std::array<secp256k1_pubkey*, 2> pointers{ &keys.front(), &keys.back() };
    return secp256k1_ec_pubkey_combine(context, &out, pointers.data(),
        pointers.size()) == ec_success && serialize(context, left, out);
```

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+static outpoint_data to_outpoint_data(const point& value) NOEXCEPT
+{
+    outpoint_data out{};
+    stream::out::fast stream{ out };
+    write::bytes::fast writer{ stream };
+    value.to_data(writer);
+    return out;
+}
+
+static bool outpoint_data_less(const outpoint_data& left,
```
evoskuil, line 71: These can all become protected class static. No change in performance, and still keeps the namespace clean, but allows for unit testing. This becomes especially important in template helpers (i.e. in hpp/ipp), because they cannot be buried in the compilation unit - even if the intent is that they be private they become public.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    const data_chunk& value) NOEXCEPT
+{
+    if (!is_compressed_key(value))
+        return false;
+
+    std::copy_n(value.begin(), ec_compressed_size, out.begin());
+    return true;
+}
+
+static ec_xonly to_xonly(const ec_uncompressed& point) NOEXCEPT
```
evoskuil, line 102: This is a general purpose util.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+    uint32_t index{};
+    for (const auto& output: outputs)
+    {
+        pay_witness_taproot_output entry{};
+        if (extract_pay_witness_taproot_output(entry, *output, index))
+            out.push_back(entry);
+
+        ++index;
+    }
```
evoskuil, line 275: Can simplify this visually:
```cpp
    uint32_t index{};
    pay_witness_taproot_output entry{};
    for (const auto& output: outputs)
        if (extract_pay_witness_taproot_output(entry, *output, index++))
            out.push_back(entry);
```

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+BC_PUSH_WARNING(NO_THROW_IN_NOEXCEPT)
+
+using chain::input;
+using chain::point;
+using chain::script;
+using chain::script_pattern;
+using chain::script_version;
+using chain::tapscript;
+using chain::transaction;
```
evoskuil, line 53: This is prob a bit too broad and can be hard to keep track of. Generally just using `chain::` namespace is sufficient.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+static constexpr auto bip341_nums_key = base16_array(
+    "50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0");
+static constexpr auto ec_odd_sign = add1(ec_even_sign);
+static constexpr size_t maximum_outputs = 2323;
+
+typedef data_array<point::serialized_size()> outpoint_data;
```
evoskuil, line 60: I don't think you need this type at all. Seems to just be inducing unnecessary copying.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+using chain::point;
+using chain::script;
+using chain::script_pattern;
+using chain::script_version;
+using chain::tapscript;
+using chain::transaction;
+
+static constexpr auto bip341_nums_key = base16_array(
+    "50929b74c1a04954b78b4b6035e97a5e078a5a0f28ec96d547bfee9ace803ac0");
+static constexpr auto ec_odd_sign = add1(ec_even_sign);
```
evoskuil, line 57: Move to secp256k1.hpp

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+static constexpr size_t maximum_outputs = 2323;
+
+typedef data_array<point::serialized_size()> outpoint_data;
+
+static outpoint_data to_outpoint_data(const point& value) NOEXCEPT
```
evoskuil, line 62: See above, don't need this.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    stream::out::fast stream{ out };
+    write::bytes::fast writer{ stream };
+    value.to_data(writer);
+    return out;
+}
+
+static bool outpoint_data_less(const outpoint_data& left,
+    const outpoint_data& right) NOEXCEPT
+{
+    return std::lexicographical_compare(left.begin(), left.end(),
```
evoskuil, line 74: Compare directly over the point struct to avoid the copy.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+        return false;
+
+    // BIP352 orders by serialized outpoints, not point's uniqueness ordering.
+    auto minimum = to_outpoint_data(inputs.front()->point());
+    for (auto it = std::next(inputs.begin()); it != inputs.end(); ++it)
+    {
+        const auto outpoint = to_outpoint_data((*it)->point());
+        if (outpoint_data_less(outpoint, minimum))
+            minimum = outpoint;
+    }
```
evoskuil, line 333: This can be reduced to std lib algo to find ref to min value.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+
+    stream::out::fast stream{ out };
+    hash::sha256t::fast<"BIP0352/Inputs"> sink{ stream };
+    sink.write_bytes(minimum);
```
evoskuil, line 337: point writes itself, done need minimum, just a ref to the point.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    ec_compressed sum{};
+    ec_secret input_hash{};
+    if (!ec_sum(sum, keys) || !compute_input_hash(input_hash, tx, sum))
+        return false;
+
+    record.prevouts_summary = sum;
+    if (!ec_multiply(record.prevouts_summary, input_hash))
+        return false;
+
+    out = std::move(record);
```
evoskuil, line 377: Drop record and just operate over out. If it's potentially large you can clear it on failure, but not critical. The caller should not generally expect a clean record on failure. But depending on the access (public/private method) you might want to clear the out value at the start if it's an accumulator, as the caller may not (e.g. in a loop).

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    return true;
+}
+
+bool scan_match::operator<(const scan_match& other) const NOEXCEPT
+{
+    return index < other.index;
+}
+
+scanner::scanner(const ec_secret& scan_secret,
+    const ec_compressed& spend_public, const std_vector<uint32_t>& labels)
```
evoskuil, line 387: label and labels could be a member of the class.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    label_tweaks_.reserve(labels_.size());
+    label_points_uncompressed_.reserve(labels_.size());
+
+    for (const auto label: labels_)
+    {
+        ec_secret tweak{};
+        ec_uncompressed point{};
+        if (!label_tweak(tweak, scan_secret_, label) ||
+            !secret_to_public(point, tweak))
+        {
```
evoskuil, line 403: Style nit, I'd drop {} here.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    if (!verify(scan_secret_) || !decompress(spend_point_, spend_public))
+        return;
+
+    label_tweaks_.reserve(labels_.size());
+    label_points_uncompressed_.reserve(labels_.size());
+
+    for (const auto label: labels_)
+    {
+        ec_secret tweak{};
+        ec_uncompressed point{};
```
evoskuil, line 400: Style nit, these declarations can go outside the loop and reduce the nesting depth. Not always better, but in this case prob so.

src/wallet/addresses/silent_payment.cpp
```diff
@@ -0,0 +1,588 @@
...
+    return true;
+}
+
+bool scanner::match_shared(bool& out,
+    const ec_compressed& shared_secret,
+    const std_vector<pay_witness_taproot_output>& outputs) const NOEXCEPT
+{
+    if (outputs.empty())
+    {
+        out = false;
```
evoskuil, line 511: move `out = false;` to first line and only change it for a true return? Otherwise it's uninitialized in a couple places.

test/crypto/elliptic_curve.cpp
```diff
@@ -167,6 +167,25 @@ BOOST_AUTO_TEST_CASE(elliptic_curve__ec_add__positive__expected)
     BOOST_REQUIRE_EQUAL(public1, public2);
 }
 
+BOOST_AUTO_TEST_CASE(elliptic_curve__ec_add_uncompressed__positive__expected)
```
evoskuil, line 170: nice readable tests.

evoskuil (changes requested):

src/chain/silent_payment.cpp
evoskuil: See updated comments.

PR #1860, opened 2026-06-05
Reject non-canonical CompactSize encodings

KY-U:
- Invalidate non canonical CompactSize encoding in `byte_reader::read_byte()`
- Fix existing tests that utilize non-canonical encodings.
- Add tests to check for invalidation of non-canonical encodings.

I have removed `byte_reader`, `bit_reader`, and `bit_flipper` 8 bytes encoding string reading tests since they were taking 1-2 min each. You can see the tested approach [here](https://github.com/KY-U/libbitcoin-system/commit/013bafe1a648affaa9a543ef5c619cf231345245).


evoskuil:
Thanks!

PR #1866, opened 2026-06-06
Add support for Testnet4 (BIP94)

thecodefactory:


src/chain/header.cpp
```diff
@@ -355,14 +355,32 @@ bool header::is_futuristic_timestamp(
     return time > future;
 }
 
+// BIP94 specific (the block's timestamp is too early on difficulty
+// adjustment block).
+bool header::is_timestamp_early(size_t height, uint32_t timestamp,
+    uint32_t previous_block_timestamp,
+    uint32_t retarget_interval) const NOEXCEPT
+{
+    static const auto max_timewarp = 600;
```
evoskuil: Move to `include\chain\enums\magic_numbers.hpp`

src/chain/header.cpp
```diff
@@ -355,14 +355,32 @@ bool header::is_futuristic_timestamp(
...
+
+    return (height % retarget_interval == 0) &&
```
evoskuil: use `is_zero(height % retarget_interval)`

src/chain/header.cpp
```diff
@@ -355,14 +355,32 @@ bool header::is_futuristic_timestamp(
...
+        (timestamp < previous_block_timestamp - max_timewarp);
```
evoskuil: use `floored_subtract(previous_block_timestamp, max_timewarp)`

The code assumes no previous block timestamp < 600, but should make this explicit.

include/bitcoin/system/error/block_error_t.hpp
```diff
@@ -42,6 +42,7 @@ enum block_error_t : uint8_t
     insufficient_block_version,
     anachronistic_timestamp,
     incorrect_proof_of_work,
+    early_timewarp_attack,
```
evoskuil: We shouldn't probably judge the cause as an attack, `early_timestamp` seems sufficient and accurate.

include/bitcoin/system/chain/header.hpp
```diff
@@ -135,6 +137,9 @@ class BC_API header
     bool is_invalid_proof_of_work(uint32_t proof_of_work_limit,
         bool scrypt=false) const NOEXCEPT;
     bool is_futuristic_timestamp(uint32_t timestamp_limit_seconds) const NOEXCEPT;
+    bool is_timestamp_early(size_t height, uint32_t timestamp,
```
evoskuil: For consistency with ` is_futuristic_timestamp()` (ad ec below) suggest ` is_early_timestamp()`.

include/bitcoin/system/chain/chain_state.hpp
```diff
@@ -185,6 +185,7 @@ class BC_API chain_state
     uint32_t minimum_block_version() const NOEXCEPT;
     uint32_t work_required() const NOEXCEPT;
     uint32_t timestamp() const NOEXCEPT;
+    uint32_t previous_block_timestamp() const NOEXCEPT;
```
evoskuil: We use `current_timestamp` and `timestamp` here since only headers/blocks have timestamps, suggest just `previous_timestamp()` here since block is implied (and the verbosity hits a lot of code).

include/bitcoin/system/chain/chain_state.hpp
```diff
@@ -239,6 +240,7 @@ class BC_API chain_state
     static uint32_t work_required_retarget(const data& values,
         const forks& forks, uint32_t proof_of_work_limit,
         uint32_t minimum_timespan, uint32_t maximum_timespan,
+        size_t retargeting_interval,
```
evoskuil: Settings must be fixed width.

include/bitcoin/system/chain/chain_state.hpp
```diff
@@ -239,6 +240,7 @@ class BC_API chain_state
...
         uint32_t retargeting_interval_seconds) NOEXCEPT;
```
evoskuil, line 249: How do `retargeting_interval` and `retargeting_interval_seconds` differ?

include/bitcoin/system/chain/header.hpp
```diff
@@ -116,7 +116,9 @@ class BC_API header
     /// Checkpoints and previous_block_hash are chain validation (not here).
 
     code check(uint32_t timestamp_limit_seconds, uint32_t proof_of_work_limit,
-        bool scrypt=false) const NOEXCEPT;
+        size_t height, uint32_t current_timestamp,
+        uint32_t previous_block_timestamp, uint32_t retarget_interval,
```
evoskuil: `retarget_interval` or `retargeting_interval`?

src/chain/chain_state.cpp
```diff
@@ -406,14 +413,42 @@ constexpr bool patch_timewarp(const forks& forks, const uint256_t& limit,
...
 uint32_t chain_state::work_required_retarget(const data& values,
     const forks& forks, uint32_t proof_of_work_limit,
     uint32_t minimum_timespan, uint32_t maximum_timespan,
+    size_t retargeting_interval,
     uint32_t retargeting_interval_seconds) NOEXCEPT
 {
     static const auto limit = compact::expand(proof_of_work_limit);
     auto target = compact::expand(bits_high(values));
 
+    if (forks.bip94)
```
evoskuil: Let's move this out to a method to try and keep this important consensus method as clean as possible.

include/bitcoin/system/chain/enums/selection.hpp
```diff
@@ -31,6 +31,7 @@ enum class selection
     none,
     mainnet,
     testnet,
```
evoskuil: Recommend another set of PRs to universally change `testnet` to `testnet3` in all contexts.

evoskuil (changes requested):
Awesome, thanks for getting this done.

src/chain/header.cpp
thecodefactory: Good catch, thanks

include/bitcoin/system/chain/chain_state.hpp
thecodefactory: https://github.com/libbitcoin/libbitcoin-system/blob/master/src/settings.cpp#L454

It looked like either way an additional parameter would be required (i.e. the value directly, or the block spacing seconds to calculate it).

include/bitcoin/system/chain/context.hpp
```diff
@@ -56,9 +57,21 @@ class BC_API context final
...
+            is_zero(height % retargeting_interval) &&
+            (timestamp <
+             floored_subtract(previous_block_timestamp, max_timewarp));
+    }
+
     /// Header context within chain.
     uint32_t flags;
     uint32_t timestamp;
+    uint32_t previous_block_timestamp;
+    uint32_t retargeting_interval;
```
evoskuil: This is config so doesn't belong in context. Context is for non-config values required for validation (generally obtained from the database).

include/bitcoin/system/chain/context.hpp
```diff
@@ -56,9 +57,21 @@ class BC_API context final
...
+        return is_enabled(chain::flags::bip94_rule) &&
+            is_zero(height % retargeting_interval) &&
+            (timestamp <
+             floored_subtract(previous_block_timestamp, max_timewarp));
+    }
+
     /// Header context within chain.
     uint32_t flags;
     uint32_t timestamp;
+    uint32_t previous_block_timestamp;
```
evoskuil: This must be added as the last field in context, as I'm pretty sure there are constructions that depended on field order.

include/bitcoin/system/chain/context.hpp
```diff
@@ -56,9 +57,21 @@ class BC_API context final
...
     }
 
+    // BIP94 specific (the block's timestamp is too early on difficulty
+    // adjustment block).
+    inline bool is_early_timestamp() const NOEXCEPT
+    {
+        return is_enabled(chain::flags::bip94_rule) &&
+            is_zero(height % retargeting_interval) &&
+            (timestamp <
+             floored_subtract(previous_block_timestamp, max_timewarp));
```
evoskuil: Please change the name `previous_block_timestamp` to `previous_timestamp` in far/field/parameter names as well.

src/chain/chain_state.cpp
```diff
@@ -368,6 +374,7 @@ uint32_t chain_state::work_required(const data& values, const forks& forks,
             settings.proof_of_work_limit,
             settings.minimum_timespan(),
             settings.maximum_timespan(),
+            settings.retargeting_interval(),
             settings.retargeting_interval_seconds);
```
evoskuil, line 386: I'll later change the name of one or both of these to avoid the obvious confusion.

evoskuil (changes requested):

include/bitcoin/system/chain/enums/flags.hpp
```diff
@@ -84,60 +84,66 @@ enum flags : uint32_t
...
+    /// -----------------------------------------------------------------------
+
+    /// Fix Satoshi's time warp bug (hard fork, security).
+    ltc_time_warp_patch = bit_right<uint32_t>(28),
+
+    /// Fix target overflow for very low difficulty (hard fork, security).
+    ltc_retarget_overflow_patch = bit_right<uint32_t>(29),
+
+    /// Use scrypt hashing for proof of work (hard fork, feature).
+    ltc_scrypt_proof_of_work = bit_right<uint32_t>(30),
```
evoskuil: Note that any time we change positions of existing bits in this enumeration it invalidates existing stores. That's unique to this specific enumeration, as these flags are stored in this bit pattern. For later upgrades we would have to map them in/out, but for now we can just break them. But to give us one more bit without breaks let's bump 26->27 ...30->31.

include/bitcoin/system/chain/context.hpp
```diff
@@ -56,13 +57,24 @@ class BC_API context final
...
     }
 
+    // Testnet4 / BIP94 specific (the block's timestamp is too early on
+    // difficulty adjustment block).
+    inline bool is_early_timestamp(uint32_t retargeting_interval) const NOEXCEPT
+    {
+        return is_enabled(chain::flags::time_warp_patch) &&
+            is_zero(height % retargeting_interval) &&
+            (timestamp <
+             floored_subtract(previous_timestamp, max_timewarp));
```
evoskuil:
```cpp        
        return is_enabled(chain::flags::time_warp_patch)
            && is_zero(height % retargeting_interval)
            && (timestamp < floored_subtract(previous_timestamp, max_timewarp));
```

evoskuil (approved):

include/bitcoin/system/chain/enums/flags.hpp
thecodefactory: Ah, right -- that's a good point.  The store should be broken also with the context member change though, so keep the proposed mappings, or update anyway?

thecodefactory: I see what you mean -- updated now :-)

PR #1871, opened 2026-06-09
Add padding byte validation during extended private key de-serialization to match bip 32's spec

KY-U:
Validate padding byte is 0 in extended private key de-serialization in order to match BIP 32's spec.

### DIFF
--- a/src/wallet/keys/hd_private.cpp
+++ b/src/wallet/keys/hd_private.cpp
@@ -145,11 +145,15 @@ hd_private hd_private::from_key(const hd_key& key, uint64_t prefixes) NOEXCEPT
     const auto parent = source.read_4_bytes_big_endian();
     const auto child = source.read_4_bytes_big_endian();
     const auto chain = source.read_hash();
-    source.skip_byte();
+    const auto padding = source.read_byte();
     const auto secret = source.read_hash();
 
     // Validate the prefix against the provided value.
     if (prefix != to_prefix(prefixes))
+        return {};
+
+    // Validate the key padding.
+    if (padding != 0x00)
         return {};
 
     const hd_lineage lineage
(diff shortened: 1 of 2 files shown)


src/wallet/keys/hd_private.cpp
```diff
@@ -145,13 +145,17 @@ hd_private hd_private::from_key(const hd_key& key, uint64_t prefixes) NOEXCEPT
     const auto parent = source.read_4_bytes_big_endian();
     const auto child = source.read_4_bytes_big_endian();
     const auto chain = source.read_hash();
-    source.skip_byte();
+    const auto padding = source.read_byte();
```
thecodefactory, line 148: It's been a long time since I've looked at this, but isn't the skip_byte or padding check only required in the hardened key case?  Quick glance I'm not seeing that distinction, so just curious if you can point me to that if it's correct.

KY-U: You're correct, the 0x00 byte used as HMAC input is only required for hardened derivation, we can see that in [`hd_private::derive_private()`](https://github.com/libbitcoin/libbitcoin-system/blob/master/src/wallet/keys/hd_private.cpp#L243-L244).

But this skip_byte() is not in derivation. It is in [parsing serialized extended private key data](https://github.com/libbitcoin/libbitcoin-system/blob/master/src/wallet/keys/hd_private.cpp#L139), where BIP32 always encodes private key data as 0x00 || ser256(k), regardless of whether the extended private key is a master key, a hardened child, or a non-hardened child.

I've updated the title to clarify that the byte-reading change relates to the serialized format and not key derivation. Thanks for pointing that out.

thecodefactory: Thank you for satisfying my curiosity, makes a lot more sense now!

thecodefactory (approved):
LGTM

PR #1874, opened 2026-06-10
Bitcoind verbose tx: total size and real witness hash (part of #778)

echennells:
First of two PRs for #778; the scriptPubKey fields (type/address/asm) follow in the second.

- **`size`**: was `serialized_size(false)` (witness-stripped); Core's `getrawtransaction` size is the witness-included total (`GetTotalSize`).

 - **`hash`**: a witness coinbase showed `null` (the BIP141 witness-commitment value from `hash(true)`). bs's `null` is the protocol-correct wtxid, but for display parity with Core/Fulcrum/electrs this adds a file-local `display_wtxid()` that recomputes the witnessed double-hash.


src/chain/json/transaction.cpp
```diff
@@ -23,12 +23,31 @@
...
 #include <bitcoin/system/stream/stream.hpp>
 
 namespace libbitcoin {
 namespace system {
 namespace chain {
 
+namespace {
+
+// wtxid for the bitcoind verbose "hash" field. hash(true) is null_hash for a
+// coinbase (BIP141 commitment leaf), but Core displays the real wtxid here.
```
evoskuil: Not actually the "real" wtxid, because the only wtxid for a coinbase is `null_hash`.

test/chain/json/transaction.cpp
```diff
@@ -113,4 +113,56 @@ BOOST_AUTO_TEST_CASE(transaction__json__conversions__expected)
     BOOST_REQUIRE(json::value_to<transaction>(value) == instance);
 }
 
+
+// bitcoind verbose serialization of a witness coinbase (mainnet block 840000),
+// checking that "size" is the witness-included total and "hash" is the
+// witnessed serialization hash (non-null), matching Core. scriptPubKey parity
```
evoskuil: Avoid the `scriptPubKey` cryptic misnomer, refer to this as the output script. Avoid the term "Core", we consistently refer to "Satoshi" for early prototype/consensus code (or "bitcoind" as you have above). If there's unfinished work lead a comment with // TODO:... Generally avoid explanations in comments if the name of the test is sufficient.

src/chain/json/transaction.cpp
```diff
@@ -23,12 +23,31 @@
...
 #include <bitcoin/system/stream/stream.hpp>
 
 namespace libbitcoin {
 namespace system {
 namespace chain {
 
+namespace {
+
+// Witnessed serialization hash for the bitcoind verbose "hash" field. The
+// wtxid of a coinbase is null_hash (BIP141), but Core displays this value.
```
evoskuil: "bitcoind"

PR #1879, opened 2026-06-12
Use the previous_timestamp from the context directly

thecodefactory:


src/chain/header.cpp
```diff
@@ -386,7 +386,7 @@ code header::accept(const context& ctx,
         return error::anachronistic_timestamp;
     if (ctx.is_invalid_work(bits_))
         return error::incorrect_proof_of_work;
-    if (ctx.is_early_timestamp(retargeting_interval))
+    if (ctx.is_early_timestamp(ctx.previous_timestamp, retargeting_interval))
```
evoskuil, line 389: Don’t pass the member to itself, just use internally.

thecodefactory: Agreed.  Thanks :sweat_smile:

PR #1905, opened 2026-07-02
Implement to_data for block and transaction views.

echennells:
`block_view::to_data(writer&, bool)` was an unimplemented stub (it asserted "not implemented"), so a `block_view` could report its serialized size but not emit its bytes; `transaction_view` had no `to_data` at all. This implements both, with witness stripping.

Witness can be downgraded but never added, mirroring `chain::transaction`: a `witness = true` call on a witnessed view writes the held bytes; otherwise the witness is stripped on write. `block_view::to_data` mirrors `serialized_size`: the witnessed form is the original buffer (a single write), otherwise it writes the header, the transaction count, and each transaction view stripped.

Round-trip tests assert the view serializes identically to the equivalent `chain::block` for witnessed, non-witness and stripped-source blocks. Full system test suite green.

### DIFF
--- a/include/bitcoin/system/chain/views/transaction_view.hpp
+++ b/include/bitcoin/system/chain/views/transaction_view.hpp
@@ -39,6 +39,11 @@ public:
     transaction_view(reader& source, const data_chunk& block_buffer,
         bool coinbase, bool witness) NOEXCEPT;
 
+    /// Serialization.
+    data_chunk to_data(bool witness) const NOEXCEPT;
+    void to_data(std::ostream& stream, bool witness) const NOEXCEPT;
+    void to_data(writer& sink, bool witness) const NOEXCEPT;
+
     /// Properties.
     bool is_valid() const NOEXCEPT;
     bool is_coinbase() const NOEXCEPT;
@@ -49,7 +54,6 @@ public:
     uint32_t version() const NOEXCEPT;
     uint32_t locktime() const NOEXCEPT;
     size_t serialized_size(bool witness) const NOEXCEPT;
-    void to_data(writer& sink, bool witness) const NOEXCEPT;
     const hash_digest& hash(bool witness) const NOEXCEPT;
 
     /// Store helpers.
--- a/include/bitcoin/system/chain/views/transaction_view.hpp
+++ b/include/bitcoin/system/chain/views/transaction_view.hpp
@@ -49,6 +49,7 @@ public:
     uint32_t version() const NOEXCEPT;
     uint32_t locktime() const NOEXCEPT;
     size_t serialized_size(bool witness) const NOEXCEPT;
+    void to_data(writer& sink, bool witness) const NOEXCEPT;
     const hash_digest& hash(bool witness) const NOEXCEPT;
 
     /// Store helpers.
--- a/src/chain/views/block_view.cpp
+++ b/src/chain/views/block_view.cpp
@@ -76,22 +76,21 @@ void block_view::to_data(std::ostream& stream, bool witness) const NOEXCEPT
     to_data(out, witness);
 }
 
-void block_view::to_data(writer& , bool ) const NOEXCEPT
+void block_view::to_data(writer& sink, bool witness) const NOEXCEPT
 {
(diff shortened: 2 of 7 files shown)


src/chain/views/block_view.cpp
```diff
@@ -76,22 +76,21 @@ void block_view::to_data(std::ostream& stream, bool witness) const NOEXCEPT
...
+    // The witnessed form is the original buffer.
+    if (witness)
+    {
+        sink.write_bytes(*buffer_);
+        return;
+    }
+
+    // Strip witness: the header and transaction count are unaffected.
+    sink.write_bytes({ buffer_->data(), std::next(buffer_->data(),
+        header::serialized_size()) });
```
evoskuil: Use 
```cpp
void write_bytes(const uint8_t* data, size_t size) NOEXCEPT override;
```

src/chain/views/transaction_view.cpp
```diff
@@ -189,6 +190,28 @@ size_t transaction_view::serialized_size(bool witness) const NOEXCEPT
     return witness && is_segregated() ? size_ : stripped_size();
 }
 
+void transaction_view::to_data(writer& sink, bool witness) const NOEXCEPT
+{
+    // Witness can be stripped but never added (mirrors chain::transaction).
+    if (witness && is_segregated())
+    {
+        sink.write_bytes({ tx_ptr_, std::next(tx_ptr_, size_) });
```
evoskuil: Use
```cpp
 void write_bytes(const uint8_t* data, size_t size) NOEXCEPT override;
```

test/chain/views/block_view.cpp
```diff
@@ -68,6 +68,105 @@ BOOST_AUTO_TEST_CASE(block_view__construct__block1a_non_witness__valid)
...
+}
+
+BOOST_AUTO_TEST_CASE(block_view__to_data__mixed_witness_and_legacy__matches_block)
+{
+    using namespace system::chain;
+
+    // One segregated (witness) transaction and one legacy (non-witness)
+    // transaction in a single block, so the per-transaction witness strip is
+    // exercised both ways within one to_data(false) call.
+    const block mixed
```
evoskuil: Move to /mocks/blocks.*

include/bitcoin/system/chain/views/transaction_view.hpp
```diff
@@ -49,6 +49,7 @@ class BC_API transaction_view final
     uint32_t version() const NOEXCEPT;
     uint32_t locktime() const NOEXCEPT;
     size_t serialized_size(bool witness) const NOEXCEPT;
+    void to_data(writer& sink, bool witness) const NOEXCEPT;
```
evoskuil: Move up to own section and add complete complement of overloads (see `block_view` and all `chain::` objects):
```cpp
    /// Serialization.
    data_chunk to_data(bool witness) const NOEXCEPT;
    void to_data(std::ostream& stream, bool witness) const NOEXCEPT;
    void to_data(writer& sink, bool witness) const NOEXCEPT;
```

src/chain/views/transaction_view.cpp
```diff
@@ -189,6 +190,28 @@ size_t transaction_view::serialized_size(bool witness) const NOEXCEPT
...
+    // and the witness data (before locktime); inputs/outputs are unchanged.
+    const auto sentinels = is_zero(witnesses_size_) ? zero : sentinels_size;
+    const auto body = std::next(tx_ptr_, version_size + sentinels);
+    const auto body_end = std::next(tx_ptr_, size_ - witnesses_size_ -
+        locktime_size);
+
+    sink.write_bytes({ tx_ptr_, std::next(tx_ptr_, version_size) });
+    sink.write_bytes({ body, body_end });
+    sink.write_bytes({ std::next(tx_ptr_, size_ - locktime_size),
+        std::next(tx_ptr_, size_) });
```
evoskuil: Same here... if you are going to do pointer and size math, use the pointer/size overload, avoiding the unnecessary intermediate `data_slice`.

Issue #1918, opened 2026-07-24
COPYING relocation in 674b8ce leaves the repository without a root license file


melvincarvalho:
The autotools reorganization in 674b8ce ("Regenerate with autotools/gnu toolchain update.", 2026-07-02) moved `COPYING` from the repository root to `builds/gnu/COPYING`. Git records it as a pure rename (`R100`) — the content is byte-identical, so this looks like an incidental side effect of grouping the GNU build files rather than an intentional licensing change.

It has some unintended consequences, though.

### GitHub no longer detects a license

```
$ gh api repos/libbitcoin/libbitcoin-system --jq .license
null
```

GitHub's license detection only inspects the repository root, so the repo currently displays no license at all — the `agpl-3.0` badge discussed back in #840 no longer resolves. Package and compliance scanners that follow the same convention will read libbitcoin-system as unlicensed.

### `COPYING` now scopes itself to `builds/gnu/`

The first line reads:

> This license applies to all works under this directory.

That sentence was accurate for the entire history of the file while it lived at the root, where "this directory" was the repository. Now, read literally, it applies the AGPL grant to the autotools build directory and not to `src/`, `include/`, or `test/`.

### The "lesser clause" no longer has a path to the source files

This is the part I think matters most, because it interacts with a deliberate past decision.

In #390, @jonassmedegaard pointed out that the source headers carried a non-standard grant referencing "additional permissions". The agreed resolution was to replace every header with the generic FSF boilerplate, which does not mention the additional permissions. That was the right call at the time — the headers pointed at a root `COPYING` that carried the §7 additional permissions, so the lesser clause reached the source through that file.

After 674b8ce, the only statement of the two §7 additional permissions lives in `builds/gnu/COPYING`, in a file whose first line disclaims applying outside its own directory, while all ~794 source headers carry boilerplate that makes no mention of them. So the permission that makes libbitcoin usable in a proprietary project — the thing the README calls the "lesser clause", and that @evoskuil explained in #1101 — currently has no textual route to `src/` or `include/`.

I don't think anyone would seriously dispute the intent; the git history and the README make it clear. It just isn't reflected in the files a recipient actually gets.

### `README.md` link is broken

README.md:13 links the license as `[COPYING](COPYING)`, which has resolved correctly since 2014 and is now a 404.

### Suggested remediation

[...]

Issue #390, opened 2016-04-25
Outdated reference to LICENSE file and libssl exception.


jonassmedegaard:
The LICENSE file was renamed 2 years ago (commit fdb87bb) and libssl exception dropped a year ago (commit 2f7bdde), yet pretty consistently both are still referenced in licensing headers.

evoskuil:
I don't see SSL referenced anywhere in this repo except for big_number.hpp/cpp test code, which has the openssl references disabled. Is this what you are referring to?

I see the reference to the file LICENSE in the copyrights. I'll change those to COPYING. The name was changed some time ago to comport to the GNU expectations of:

- INSTALL
- NEWS
- README
- COPYING
- AUTHORS
- ChangeLog

evoskuil:
LICENSE has been replaced with COPYING in the copyright headers.

jonassmedegaard:
Quoting Eric Voskuil (2016-04-25 10:34:39)
> I don't see SSL referenced anywhere in this repo except for 
> big_number.hpp/cpp test code, which has the openssl references 
> disabled. Is this what you are referring to?

What I am referring to is the difference between generic boilerplate 
license grant as recommended at the top of your COPYING file:

> This program is free software: you can redistribute it and/or modify 
> it under the terms of the GNU Affero General Public License as 
> published by the Free Software Foundation, either version 3 of the 
> License, or (at your option) any later version.

...and the actual license grant used throughout the code:

> libbitcoin is free software: you can redistribute it and/or modify it 
> under the terms of the GNU Affero General Public License with 
> additional permissions to the one published by the Free Software 
> Foundation, either version 3 of the License, or (at your option) any 
> later version. For more information see LICENSE.

In particular the "with additional permissions" part.


I recommend to use the generic boilerplate verbatim (not even mention 
"libbitcoin" in that paragraph, just the generic "This program").


 - Jonas

-- 
 * Jonas Smedegaard - idealist & Internet-arkitekt
 * Tlf.: +45 40843136  Website: http://dr.jones.dk/

 [x] quote me freely  [ ] ask before reusing  [ ] keep private

evoskuil:
And regarding SSL?

evoskuil:
So change this:

```cpp
/**
 * Copyright (c) 2011-2015 libbitcoin developers (see AUTHORS)
 *
 * This file is part of libbitcoin.
 *
 * libbitcoin is free software: you can redistribute it and/or modify
[...]

PR #1919, opened 2026-07-24
Restore COPYING at repository root

melvincarvalho:
Addresses the first point of #1918.

674b8ce ("Regenerate with autotools/gnu toolchain update.") moved `COPYING` from the repository root into `builds/gnu/` alongside the other GNU boilerplate. Git records it as a pure rename (`R100`) with byte-identical content, so it reads as collateral from grouping the build files rather than an intentional change.

The side effect is that the repository presents no license file at its root:

```
$ gh api repos/libbitcoin/libbitcoin-system --jq .license
null
```

GitHub's detection only inspects the root, and `README.md:13` links the license as `[COPYING](COPYING)`, which now 404s.

This PR adds a byte-identical copy back at the root. `builds/gnu/COPYING` is deliberately retained rather than moved, because `builds/gnu/configure.ac` uses `AM_INIT_AUTOMAKE([subdir-objects])` without `[foreign]` — which requires `COPYING` next to `configure.ac` — and `builds/gnu/Makefile.am` installs it via `doc_DATA`.

```
$ cmp builds/gnu/COPYING COPYING && echo identical
identical
```

Worth noting: restoring the file to the root also makes its own first line — "This license applies to all works under this directory" — accurate again, as it was from 2014 until 674b8ce. No wording change is needed for that.

**Scope.** This is intentionally the mechanical part only: one added file, nothing modified, no license text authored. The remaining points in #1918 — whether to reword that first line so it isn't directory-relative, whether to state explicitly that the §7 additional permissions apply repo-wide (relevant because #390 removed that language from the source headers), the `AUTHORS` pointer, and the missing LGPL-2.1 text for the vendored qrencode — are all editorial or independent, and I've left them for you to decide there rather than presuming on them here.

Happy to close this if you'd rather handle the relocation in the toolchain that produced 674b8ce.

### DIFF
--- a/COPYING
+++ b/COPYING
@@ -1,0 +1,652 @@
+This license applies to all works under this directory.
+
+Libbitcoin - C++ Bitcoin Development Toolkit and Applications
+Copyright © 2011-2019 libbitcoin developers (see AUTHORS)
+
+This program is free software: you can redistribute it and/or modify
+it under the terms of the GNU Affero General Public License as published by
+the Free Software Foundation, either version 3 of the License, or
+(at your option) any later version.
+
+This program is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU Affero General Public License for more details.
+
+You should have received a copy of the GNU Affero General Public License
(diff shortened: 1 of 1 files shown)


evoskuil:
> The side effect is that the repository presents no license file at its root

This is of no consequence.

PR #1927, opened 2026-08-04
fix: verify taproot commitment for all leaf versions

brunoerg:
extract_taproot() branched on the control block leaf version before checking the commitment to the output key, so a script path spend with any leaf version other than 0xc0 returned success without proving any commitment at all. BIP341 orders these the other way around: the tapleaf hash, the merkle path and the "fail if q != x(Q) or c[0] & 1 != y(Q) mod 2" test all precede the leaf version branch, as in Core's VerifyWitnessProgram.

Every 32-byte v1 program was therefore spendable by anyone, with the witness <anything> <anything> <0xc2 || 32 arbitrary bytes>. This accepts invalid blocks rather than rejecting valid ones, so it is chain splitting.

Compute the tapleaf hash and verify the commitment for every leaf version, then branch: 0xc0 executes the tapscript, other versions remain unencumbered as before.

### DIFF
--- a/src/chain/witness_extract.cpp
+++ b/src/chain/witness_extract.cpp
@@ -194,21 +194,24 @@ code witness::extract_taproot(hash_cptr& out_leaf, script::cptr& out_script,
             if (!control.is_valid())
                 return error::invalid_witness;
 
+            const auto& key = unsafe_array_cast<uint8_t, ec_xonly_size>(
+                program->data());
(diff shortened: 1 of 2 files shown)


brunoerg:
Thanks @evoskuil for adding tests.

evoskuil:
Thanks for the PR!

PR #1929, opened 2026-08-04
fix: 3 tapscript consensus divergences from BIP341/342

brunoerg:
Follow-up to #1927, which fixed the missing taproot commitment check for unknown leaf versions. The same review surfaced three further divergences between tapscript execution and BIP341/342. Each is a separate commit with tests (asked an LLM to help with the tests btw).

1. Sigop budget derived from the wrong witness
2. `OP_CHECKSIGADD` parsed the endorsement for unknown public key types
3. Signature hash cache not invalidated when the subscript changes

### DIFF
--- a/include/bitcoin/system/impl/machine/interpreter.ipp
+++ b/include/bitcoin/system/impl/machine/interpreter.ipp
@@ -1162,6 +1162,10 @@ op_check_multisig_verify() NOEXCEPT
     const auto subscript = this->subscript(endorsements);
     const auto bip66 = this->is_enabled(flags::bip66_rule);
 
+    // This subscript is stripped of these endorsements, so a hash cached by a
+    // previous signature op (with a different subscript) cannot be reused.
+    this->uncache();
+
     // Keys may be empty.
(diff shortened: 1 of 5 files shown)


evoskuil:
One style nit on the tests, we don't enforce the 80 char rule on test lines, and don't allow parameter wrapping (or conditions) in tests. Test lines can be shortened with intermediate vars.

You get credit for the very first interpreter unit tests cases. Which means you can delete the stub:

```cpp
BOOST_AUTO_TEST_CASE(interpreter_test)
{
    BOOST_REQUIRE(true);
}
```

brunoerg:
> One style nit on the tests, we don't enforce the 80 char rule on test lines, and don't allow parameter wrapping (or conditions) in tests. Test lines can be shortened with intermediate vars.
> 
> You get credit for the very first interpreter unit tests cases. Which means you can delete the stub:
> 
> ```c++
> BOOST_AUTO_TEST_CASE(interpreter_test)
> {
>     BOOST_REQUIRE(true);
> }
> ```

Done

PR #1932, opened 2026-08-06
Require minimal OP_IF/OP_NOTIF argument under tapscript

brunoerg:
BIP342 requires that the argument to OP_IF and OP_NOTIF be exactly the
empty vector or [0x01]: "fail if the top stack element is not exactly
an empty vector or the one byte vector 0x01". 

boolean::from_chunk(bool&, ...) tested only vary.front(), so the size
and every byte beyond the first were ignored. Any chunk with a leading
0x01 passed as a minimal true, including [0x01, 0x00] and
[0x01, 0x00, 0x00, 0x00], and any chunk with a leading 0x00 passed as a
minimal false, including [0x00], which bip342 admits only as [].
front() is also the low order byte rather than the value, so a chunk
such as [0x00, 0x01] was evaluated as false where it is true.

This reaches consensus through peek_minimal_bool() and pop_bool_(), so
tapscripts whose OP_IF argument bip342 rejects were executed, and the
branch taken could differ from the one bip342 specifies. It accepts
what must be rejected, so it is chain splitting.

Evaluate the boolean over the full chunk and report minimality
separately, as the size one and value one test that bip342 defines.

### DIFF
--- a/include/bitcoin/system/impl/machine/number_boolean.ipp
+++ b/include/bitcoin/system/impl/machine/number_boolean.ipp
@@ -78,7 +78,11 @@ from_chunk(bool& value, const data_chunk& vary) NOEXCEPT
         return true;
     }
 
-    return from_integer(value, vary.front());
+    // A non-minimal chunk retains its (non-minimal) boolean interpretation.
+    value = from_chunk(vary);
+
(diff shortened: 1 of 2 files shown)


evoskuil:
Thanks again @brunoerg and for the tests!

PR #1935, opened 2026-08-10
Require canonical scriptSig push for p2sh-wrapped witness spends

brunoerg:
The scriptSig of a p2sh-wrapped witness spend must be byte-exactly a
single nominal push of the embedded script. Only the operation count was
checked, and is_relaxed_push_pattern admits push_one_size/two/four, so a
scriptSig of [4c 16 <22-byte program>] was accepted where Core returns
SCRIPT_ERR_WITNESS_MALLEATED_P2SH. The p2sh hash is unchanged and bip143
does not commit to the scriptSig, so this permitted third party txid
malleation of otherwise valid transactions.


evoskuil:
Thanks!

